PR #3, opened 2019-11-08
Build fix for project.

craigdbarber:
Added build precommit hook to package.json, updated action.yml to point to self-contained output file, removed intermediate output files from dist.


craigdbarber:
@imjohnbo PTAL

PR #30, opened 2019-12-27
add-GOOGLE_APPLICATION_CREDENTIALS

brachi-wernick:
Add option to set specific name to ADC file, in order to be able set it later in env variables.
(updated also readme)


brachi-wernick:
@googlebot I fixed it.

brachipa:
@googlebot I fixed it.

brachi-wernick:
@googlebot I fixed it

setup-gcloud/README.md
```diff
@@ -53,4 +53,25 @@ steps:
...
+steps:
+    - uses: actions/checkout@v1
+    - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+      with:
+       version: '270.0.0'
+       service_account_email: ${{ secrets.GCP_SA_EMAIL }}
+       service_account_key: ${{ secrets.GCP_SA_KEY }}
+       service_account_file_name: service_account
+    - name: add gcp cradentilas
+      run:  echo ::set-env name=GOOGLE_APPLICATION_CREDENTIALS::/tmp/service_account
```
EronWright: Consider taking the whole path as a parameter rather than just the file name, to improve the symmetry between the `setup-gcloud` step and subsequent steps.  i.e.

Also, replace `cradentilas` with `credentials`.

```
    - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
      with:
       version: '270.0.0'
       service_account_email: ${{ secrets.GCP_SA_EMAIL }}
       service_account_key: ${{ secrets.GCP_SA_KEY }}
       service_account_file_name: /tmp/service_account
    - name: add gcp credentials
      run:  echo ::set-env name=GOOGLE_APPLICATION_CREDENTIALS::/tmp/service_account
```

brachi-wernick: @EronWright done!

setup-gcloud/README.md
```diff
@@ -53,4 +53,25 @@ steps:
...
 
-* `service_account_key`: The service account key which will be used for authentication. This key should be [created](https://cloud.google.com/iam/docs/creating-managing-service-account-keys), encoded as a [Base64](https://en.wikipedia.org/wiki/Base64) string (eg. `cat my-key.json | base64` on macOS), and stored as a [secret](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets). 
+* `service_account_key`: The service account key which will be used for authentication. This key should be [created](https://cloud.google.com/iam/docs/creating-managing-service-account-keys), encoded as a [Base64](https://en.wikipedia.org/wiki/Base64) string (eg. `cat my-key.json | base64` on macOS), and stored as a [secret](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets).
+
+* `service_account_file_name`: (Optional) The name of the file that the service account ADC will be saved into, this is used when you need to reference this file with `GOOGLE_APPLICATION_CREDENTIALS`, if not specify, name will be randomize.
```
craigdbarber: typo: s/specify/specified

setup-gcloud/README.md
```diff
@@ -53,4 +53,25 @@ steps:
...
+steps:
+    - uses: actions/checkout@v1
+    - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+      with:
+       version: '270.0.0'
+       service_account_email: ${{ secrets.GCP_SA_EMAIL }}
+       service_account_key: ${{ secrets.GCP_SA_KEY }}
+       service_account_file_name: /tmp/service_account
+    - name: add gcp credentilas
+      run:  echo ::set-env name=GOOGLE_APPLICATION_CREDENTIALS::/tmp/service_account
```
craigdbarber: Rather than having the user have to include this as a manual step, we can do this for them within the action's code, see: https://github.com/actions/toolkit/tree/master/packages/core#exporting-variables

setup-gcloud/src/setup-gcloud.ts
```diff
@@ -49,15 +49,27 @@ async function run() {
...
         }
         resolve(path);
       });
     });
-    await fs.writeFile(tmpKeyFilePath, Base64.decode(serviceAccountKey));
+
+    const serviceAccountFileName = core.getInput('service_account_file_name');
+
+    if (serviceAccountFileName) {
+      tmpKeyFilePath = `/tmp/${serviceAccountFileName}`;
```
craigdbarber: If the parameter is meant to be the full file path as per @EronWright 's suggestoin, then that needs to be reflected here.

setup-gcloud/src/setup-gcloud.ts
```diff
@@ -49,15 +49,27 @@ async function run() {
...
+    }
+
+    if (!fs.existsSync(tmpKeyFilePath)) {
```
craigdbarber: What about the case when the file does exist? Presumably we'd still want to write the key to the file in-case the key has been changed in the secret store.

setup-gcloud/src/setup-gcloud.ts
```diff
@@ -49,15 +49,27 @@ async function run() {
     }
 
     // write the service account key to a temporary file
-    const tmpKeyFilePath = await new Promise<string>((resolve, reject) => {
+    let tmpKeyFilePath = await new Promise<string>((resolve, reject) => {
```
craigdbarber: I'd rather not overload the purpose of this variable. This variable is distinctly meant to represent a temporary file. In the case where a use wants a persistent file, we should use a separate variable to represent that.

setup-gcloud/README.md
craigdbarber: Also, please add a note making clear that this will cause a key file to persist on the runner's environment.

craigdbarber:
Thanks very much for taking the time to add this behavior! A couple of in-line comments.

craigdbarber:
@brachi-wernick it looks like our CLA scanner bot has picked up commits with an email address which doesn't have the signed CLA. Could you please re-base your commits to use the email that is signed? Thanks very much :)

setup-gcloud/src/setup-gcloud.ts
brachipa: sure, forget this.

setup-gcloud/README.md
brachipa: very nice, I'll do that

setup-gcloud/src/setup-gcloud.ts
brachipa: yes but later this variable is being used :
https://github.com/GoogleCloudPlatform/github-actions/blob/67a97a91f262da1798e5d0ec09b38057e89baf32/setup-gcloud/src/setup-gcloud.ts#L69
and
https://github.com/GoogleCloudPlatform/github-actions/blob/67a97a91f262da1798e5d0ec09b38057e89baf32/setup-gcloud/src/setup-gcloud.ts#L76

and I don't want to duplicate this logic for 2 different variables.
Also, IMO both representing the same thing, a file path, just it is being defined by 2 options: random or by user input

brachipa:
> @brachi-wernick it looks like our CLA scanner bot has picked up commits with an email address which doesn't have the signed CLA. Could you please re-base your commits to use the email that is signed? Thanks very much :)

I sign with both accounts, both are mine. I don't know why it still asking me, any idea?

setup-gcloud/src/setup-gcloud.ts
brachipa: correct, I'll remove this check

setup-gcloud/src/setup-gcloud.ts
craigdbarber: Fair point :)

craigdbarber:
> > @brachi-wernick it looks like our CLA scanner bot has picked up commits with an email address which doesn't have the signed CLA. Could you please re-base your commits to use the email that is signed? Thanks very much :)
> 
> I sign with both accounts, both are mine. I don't know why it still asking me, any idea?

The bot's output is saying that the ***@moonactive.com email address has not signed.

brachipa:
when I log in with the `@moonactive` account I see that I signed in Dec 26, 2019 16:37 PST.

craigdbarber:
> when I log in with the `@moonactive` account I see that I signed in Dec 26, 2019 16:37 PST.

@brachi-wernick try posting in a new comment: "@googlebot I consent"

brachipa:
@googlebot I consent

brachi-wernick:
@googlebot I consent

setup-gcloud/README.md
```diff
@@ -54,4 +54,10 @@ steps:
 
 * `service_account_email`: (Optional) The service account email which will be used for authentication.
 
-* `service_account_key`: (Optional) The service account key which will be used for authentication. This key should be [created](https://cloud.google.com/iam/docs/creating-managing-service-account-keys), encoded as a [Base64](https://en.wikipedia.org/wiki/Base64) string (eg. `cat my-key.json | base64` on macOS), and stored as a [secret](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets). 
+* `service_account_key`: The service account key which will be used for authentication. This key should be [created](https://cloud.google.com/iam/docs/creating-managing-service-account-keys), encoded as a [Base64](https://en.wikipedia.org/wiki/Base64) string (eg. `cat my-key.json | base64` on macOS), and stored as a [secret](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets). 
+
+* `service_account_file_path`: (Optional) Full file path, where the service account ADC will be saved into, if not specified, name will be randomize.
```
EronWright: May I suggest that if `service_account_file_path` is not set, then `GOOGLE_APPLICATION_CREDENTIALS` be not set.

brachi-wernick:
@craigdbarber I can't see how #53 solve this issue.

Issue #17, opened 2019-12-10
Add a "build and deploy to Cloud Run" example workflow


steren:
GitHub Actions are an ideal way to automatically build and deploy to [Cloud Run](https://cloud.run).

I suggest we add an [example workflow](https://github.com/GoogleCloudPlatform/github-actions/tree/master/example-workflows)

There already are a lot of such third party tutorials:
  - [Deploying projects to Cloud Run using GitHub Actions](https://misfra.me/2019/09/deploying-projects-to-cloud-run-using-github-actions/)
  - [Publish your Cloud Run App with GitHub Actions](https://medium.com/better-programming/publish-your-cloud-run-app-with-github-actions-6c18ff5c5ee4)
  -  [Automate Cloud Run deployment in a minute (with GitHub actions)](https://medium.com/@ujwaldhakal/automate-cloud-run-deployment-in-a-minute-cb85e7db9f82)
  - [Github Action for Google Cloud Run](https://github.com/marketplace/actions/cloud-run)

craigdbarber:
Thanks @steren for the feedback. We'll incorporate this information into our roadmap planning process. In the meantime, feel free to submit a PR for any new example workflows you'd like to see.

pollosp:
```
name: Deploy the application to Google Cloud Run
on:
  push:
    branches:
      - 'master'

jobs:
  setup-build-publish-deploy:
    name: Setup, Build, Publish, and Deploy
    runs-on: ubuntu-latest
    steps:
    - name: Checkout
      uses: actions/checkout@master

    # Setup gcloud CLI
    - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
      with:
        version: '270.0.0'
        service_account_key: ${{ secrets.GCLOUD_AUTH }}

    # Configure docker to use the gcloud command-line tool as a credential helper
    - run: |
        # Set up docker to authenticate
        # via gcloud command-line tool.
        gcloud auth configure-docker
      
    # Build the Docker image
    - name: Build
      run: |        
        docker build . --tag eu.gcr.io/${{ secrets.GCLOUD_PROJECT }}/${{ secrets.GCLOUD_APP_NAME }}
    # Push the Docker image to Google Container Registry
    - name: Publish
      run: |
        docker push eu.gcr.io/${{ secrets.GCLOUD_PROJECT }}/${{ secrets.GCLOUD_APP_NAME }}
        
    # Deploy the Docker image to the GKE cluster
[...]

PR #36, opened 2020-01-14
Add Cloud run Example Workflow

averikitsch:


example-workflows/cloud-run/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+  * `RUN_REGION`: the region in which the resource will be deployed.
+
+  * `SERVICE_NAME`: your preferred name of your service and image.
+
+1. [Add the the following role to your service account][roles]:
+
+  * `Cloud Run Admin`
+  * `Cloud Build Editor`
+  * `Viewer`
```
glasnt: Which Viewer? Cloud Build? Cloud Run?

example-workflows/cloud-run/cloud-run.yml
```diff
@@ -0,0 +1,61 @@
...
+        run: |
+          gcloud builds submit -t gcr.io/$PROJECT_ID/$SERVICE_NAME:$GITHUB_SHA
+
+      # Deploy image to Cloud Run
+      - name: Deploy
+        run: |
+          gcloud run deploy $SERVICE_NAME \
+            --image gcr.io/$PROJECT_ID/$SERVICE_NAME:$GITHUB_SHA \
+            --platform managed \
+            --allow-unauthenticated
```
grayside: Generally recommended to not include `--allow-unauthenticated` in your CD script.

Some considerations which may not apply to everyone's use case or testing constraints (including "easy of learning from an example")

* Requires broader privileges than regular Cloud Run deployment
* Removes the go/no-go decision from humans
* Locking down access is one step you might want to take in an emergency situation (e.g., data corruption), but this configuration will re-expose the service automatically.

If you do remove it, Cloud Run will persist this setting from revision to revision if it's used just a single time.

example-workflows/cloud-run/README.md
averikitsch: Project Viewer

glasnt:
```suggestion
  * `Viewer` (roles/viewer)
```

Also I'd add `roles/run.admin` and `roles/cloudbuild.builds.editor`; these are the role IDs that the CLI needs, and would help with any miscommunication when the role titles require disambiguation.

example-workflows/cloud-run/cloud-run.yml
craigdbarber: Perhaps we should strive for consistency with the official getting started docs for this project on this matter: https://cloud.google.com/run/docs/deploying

craigdbarber:
Looks awesome, thanks for adding this! A few comments in-line.

craigdbarber:
Fixes #17

example-workflows/cloud-run/cloud-run.yml
grayside: Deploying page does not state an opinion about the use of --allow-unauthenticated, and https://cloud.google.com/run/docs/continuous-deployment-with-cloud-build deliberately leaves it out for reasons of point 2.

example-workflows/cloud-run/README.md
grayside: I'm not quite sure why you haven't run into the Service Account user role necessity here, maybe because Cloud Build is somehow acting as an intermediary?

[https://cloud.google.com/run/docs/continuous-deployment-with-cloud-build#continuous-iam](Continuous deployment with minimal IAM permissions) from the official docs explains this role is needed to ensure you can't deploy code with privileges which exceed those of the deployer.

example-workflows/cloud-run/cloud-run.yml
craigdbarber: If you click on the "Command line" tab you'll see it's directly mentioned. @averikitsch would it be possible to follow suit and include that snippet in the README for this?

example-workflows/cloud-run/README.md
averikitsch: I've changed the permissions based on Katie's and I conversation (https://gist.github.com/glasnt/e66f944ae4136ad410ddba41a34f5bd0). I'm not basing this PR on the Cloud Build docs because those use Cloud Build to handle all aspects of building and deployment to Cloud Run which isn't what we are doing here.

example-workflows/cloud-run/cloud-run.yml
averikitsch: I've removed the flag and added a note about how it works, how to find more information, and  how to change it.

PR #38, opened 2020-01-18
fix using installGcloudSDK()

tell-y:
The line which is written installGcloudSDK() does not have `await` operator.
So I fixed it.


tell-y:
@googlebot I signed it!

tell-y:
Could you review this?
@craigdbarber

Issue #12, opened 2019-12-01
Bypass authentication in setup-gcloud


ynouri:
Hi,

I'm using the setup-gcloud Github action to setup a [Cloud Pub/Sub Emulator](https://cloud.google.com/pubsub/docs/emulator) in my workflow, for which I theoretically don't need to authenticate. I would like to avoid creating a dummy service account just for that purpose.

Is there a way to bypass the authentication part of setup-gcloud? I tried to pass a dummy key but I get the following error:

```
Run GoogleCloudPlatform/github-actions/setup-gcloud@master16s
##[error]Node run failed with exit code 1
Run GoogleCloudPlatform/github-actions/setup-gcloud@master
gcloud auth activate-service-account --key-file=/tmp/tmp-2698ACb7IjACbz1P
/bin/tar xz -C /home/runner/work/_temp/1af810e3-a86c-4475-a434-53d924a4922d -f /home/runner/work/_temp/8f778557-1f43-4cab-b7b5-e77550f7f01c
ERROR: (gcloud.auth.activate-service-account) Missing required argument [ACCOUNT]: An account is required when using .p12 keys
```

Maybe a setting making the authentication optional would work?

Thanks!

24cafeteria:
<img src="https://yyglsa.by.files.1drv.com/y4mwt1RyVlWM8qufCd7k4a-68ZwwsJdYzeFJGcMqkdIeYZMoyZgxZJnl5vkF0cZo69EBggBU9msKYuJrUYAVzpO3OJ6yVHiLXpWVkkTZWJD4Qjd3sIKDlSkGYA95giE7uX8Anb6pIGsf3RuEW4w0-Gx75_-VTespT8LLhqdfXI6rJl9eHcmtanwYRBt2RLeI5Wpnvr4JiSbyXmfxB6XxmtzWw?width=606&height=606&cropmode=none" width="606" height="606" />

24cafeteria:
#11

24cafeteria:
Run GoogleCloudPlatform/github-actions/setup-gcloud@master16s
##[error]Node run failed with exit code 1
Run GoogleCloudPlatform/github-actions/setup-gcloud@master
gcloud auth activate-service-account --key-file=/tmp/tmp-2698ACb7IjACbz1P
[...]

PR #43, opened 2020-01-28
Made authentication optional

craigdbarber:
Addresses #12


craigdbarber:
Ignore the macos timeout. I've created an issue with github-actions/toolkit to track the perf issues we're encountering with the macos runners: https://github.com/actions/toolkit/issues/329

craigdbarber:
> Is it possible to surface a log that gcloud is being used unauthenticated?

Good call! Done.

Issue #31, opened 2020-01-09
Ability to specify latest gcloud version


brettski:
While setting up the action you specify the version of gcloud to use, E.g. `174.0.0.1`. Have the ability to specify latest version. 

### Example
```yml
steps:
- uses: actions/checkout@v1
- uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
  with:
      version: 'latest'
      service_account_email: ${{ secrets.GCP_SA_EMAIL }}
      service_account_key: ${{ secrets.GCP_SA_KEY }}
- run: gcloud info
```

If changing major versions may not be desirable without prior testing, perhaps a semver-type formatting to run at any patched or minor versions of the major version specified.

example: `version: '^274.0.0'`

PR #45, opened 2020-01-30
Add support for specifying `latest` for version

craigdbarber:
Integration tests successfully passed: https://github.com/GoogleCloudPlatform/github-actions/actions/runs/32798521

Addresses: https://github.com/GoogleCloudPlatform/github-actions/issues/31


setup-gcloud/src/version-util.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+import * as httpm from 'typed-rest-client/HttpClient';
+
+/**
+ * @returns The latest stable version of the gcloud SDK.
+ */
+export async function getLatestGcloudSDKVersion(): Promise<string> {
+  const client: httpm.HttpClient = new httpm.HttpClient(
+    'github-actions-setup-gcloud',
```
averikitsch: Can we pass this in from the other fie where it is defined?

setup-gcloud/src/version-util.ts
```diff
@@ -0,0 +1,52 @@
...
+  );
+  return client
+    .get(
+      'https://api.github.com/repos/GoogleCloudPlatform/cloud-sdk-docker/tags',
+    )
+    .then(res => {
```
averikitsch: Can we change this to async/await pattern to stay consistent?

setup-gcloud/src/version-util.ts
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
craigdbarber: You caught me copying and pasting :)
Done.

setup-gcloud/src/version-util.ts
craigdbarber: Done.

setup-gcloud/__tests__/version-util.test.ts
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
craigdbarber: Done.

setup-gcloud/src/version-util.ts
craigdbarber: I would love to, sadly the library we're using here doesn't support it, so we have to wrap around it :)

averikitsch: I'm seeing async/await in their "docs" aka tests (https://github.com/microsoft/typed-rest-client/blob/master/test/tests/httptests.ts#L52). This code is just getting harder and harder to read.

craigdbarber: Oh snap! You are correct. For some reason when I first looked into using the library I had decided it didn't support it. Glad to see that's not the case. Good catch!

Issue #50, opened 2020-02-04
Feature request: Google Secret Manager example/support


thisguychris:
Hi, I just recently found out about this, I hope we can add an example in this repo. Especially when Github Actions doesn't support org-level secrets yet.

Some inspiration from azure key vault:

```yaml
on: [push]

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      # checkout the repo
    - uses: actions/checkout@master
    - uses: Azure/login@v1
      with:
        creds: ${{ secrets.AZURE_CREDENTIALS }} # Define secret variable in repository settings as per action documentation
    - uses: Azure/get-keyvault-secrets@v1.0
      with:
        keyvault: "myKeyVault"
        secrets: 'mySecret1, mySecret2'
      id: myGetSecretAction
    - uses: Azure/docker-login@v1
      with:
        login-server: mycontainer.azurecr.io
        username: ${{ steps.myGetSecretAction.outputs.mySecret1 }}
        password: ${{ steps.myGetSecretAction.outputs.mySecret2 }}
    - run: |
        cd go-sample
        docker build . -t my.azurecr.io/myimage:${{ github.sha }}
        docker push my.azurecr.io/myimage:${{ github.sha }}
        cd ..
```

Where you can get the outputs via steps?

cc @sethvargo

sethvargo:
Yup, it's on my list and in a half-baked POC on my laptop 😄. There are higher priority integrations at the moment, but it's on my roadmap.

thisguychris:
woohoo 🙌 . @sethvargo is it something like how azure key vault does it? Using outputs of a step. I think that's a really intuitive way of using it.

sethvargo:
That was my original plan, but there's also a new actions API that I haven't investigated yet.

thisguychris:
@sethvargo is there a way to use secret manager using an API key instead of email and key combination? I'm asking since Github doesn't have org level secrets yet, 1 secret versus 2 secret is a big deal if you are dealing with hundreds of repos.

sethvargo:
You can access the API using any supported Google Auth (service account key or short-lived identity token)

PR #53, opened 2020-02-13
Add Secret Manager action

sethvargo:
This is an initial POC for an action that fetches secrets from [Secret Manager](https://cloud.google.com/secret-manager). It currently works when credentials are passed directly to the action, but I cannot get it to work with the `setup-gcloud` action - it does not find the ADC. 

We also need to create a GitHub Secret for the tests to run successfully.

/cc @craigdbarber 

Fixes GH-50
Closes GH-30


.github/workflows/get-secretmanager-secrets-it.yml
```diff
@@ -0,0 +1,47 @@
+name: get-secretmanager-secrets Integration
+on: [push]
+jobs:
+  gcloud:
+    name: with setup-gcloud
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: ./setup-gcloud
```
sethvargo: I cannot get this to work. I'm not sure what's going on, but SM doesn't find the ADC. Weirdly, it works fine running locally with gcloud installed, so I think it's something in the action.

.github/workflows/get-secretmanager-secrets-it.yml
```diff
@@ -0,0 +1,47 @@
...
+      with:
+        service_account_email: ${{ secrets.SETUP_GCLOUD_IT_EMAIL }}
+        service_account_key: ${{ secrets.SETUP_GCLOUD_IT_KEY_B64 }}
```
sethvargo: In order to properly test this, we need an envvar that has a b64-encoded service account json and a non-encoded one.

.github/workflows/get-secretmanager-secrets-it.yml
```diff
@@ -0,0 +1,47 @@
...
+    name: with base64 json creds
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - id: secrets
+      uses: ./get-secretmanager-secrets
+      with:
+        credentials: ${{ secrets.SETUP_GCLOUD_IT_KEY_B64 }}
+        secrets: |-
+          token:${{ secrets.GET_SECRETMANAGER_SECRET_REF }}
```
sethvargo: Need to create this GitHub secret. That will involve also creating a secret in Secret Manager. Not sure if you have a GCP project that you're already using for tests.

.github/workflows/setup-gcloud-it.yml
```diff
@@ -1,4 +1,4 @@
-name: Integration Tests
+name: setup-gcloud Integration
```
sethvargo: Renamed these

get-secretmanager-secrets/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+
+## Authorization
+
+There are a few ways to authenticate this action. The caller must have
+permissions to access the secrets being requested.
+
+### Via the setup-gcloud action
+
+You can provide credentials using the [setup-gcloud][setup-gcloud] action:
```
thisguychris: @sethvargo, is this a typo? I see two `[setup-gcloud][setup-gcloud]`; perhaps you were aiming for a link instead?

.github/workflows/get-secretmanager-secrets-it.yml
craigdbarber: This is because the setup-gcloud action doesn't create the ADC. There a PR open for adding this which has stalled, feel free to borrow inspiration from it and add it into this PR: https://github.com/GoogleCloudPlatform/github-actions/pull/30

.github/workflows/get-secretmanager-secrets-it.yml
craigdbarber: We do, will send info for it offline.

.github/workflows/get-secretmanager-secrets-it.yml
```diff
@@ -0,0 +1,47 @@
...
+      with:
+        service_account_email: ${{ secrets.SETUP_GCLOUD_IT_EMAIL }}
+        service_account_key: ${{ secrets.SETUP_GCLOUD_IT_KEY_B64 }}
+    - id: secrets
+      uses: ./get-secretmanager-secrets
+      with:
+        secrets: |-
+          token:${{ secrets.GET_SECRETMANAGER_SECRET_REF }}
+    - name: get output
+      run: echo '${{ steps.secrets.outputs.token }}'
```
craigdbarber: Do we want to be printing a secret to the build log?

.github/workflows/get-secretmanager-secrets.yml
```diff
@@ -0,0 +1,21 @@
...
+
+    - uses: actions/setup-node@master
+      with:
+        node-version: 12.x
+
+    - run: npm install
+      working-directory: ./get-secretmanager-secrets
+
+    - run: npm test
+      working-directory: ./get-secretmanager-secrets
```
craigdbarber: Also suggest including a lint check similar to the setup-gcloud action: https://github.com/GoogleCloudPlatform/github-actions/blob/9ef0382a75df2efcc72850f4e963927a45d59414/.github/workflows/setup-gcloud.yml#L23

.github/workflows/get-secretmanager-secrets.yml
```diff
@@ -0,0 +1,21 @@
...
+      matrix:
+        operating-system: [ubuntu-latest, windows-latest, macos-latest]
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/setup-node@master
+      with:
+        node-version: 12.x
+
+    - run: npm install
```
craigdbarber: Please include step names to be consistent with the the setup-gcloud workflow.

get-secretmanager-secrets/.gitignore
```diff
@@ -0,0 +1,5 @@
+node_modules/
```
craigdbarber: Please copy the .gitignore from the setup-gcloud action: https://github.com/GoogleCloudPlatform/github-actions/blob/master/setup-gcloud/.gitignore

get-secretmanager-secrets/README.md
```diff
@@ -0,0 +1,134 @@
+# get-secretmanager-secrets
```
craigdbarber: Add missing license header.

get-secretmanager-secrets/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+This action fetches secrets from [Secret Manager][sm] and makes then available
```
craigdbarber: s/then/them

get-secretmanager-secrets/README.md
```diff
@@ -0,0 +1,134 @@
...
+    # Short form - "latest" version
+    <project-id>/<secret-id>
+    ```
+
+-   `credentials`: (Optional) [Google Service Account JSON][sa] credentials,
+    typically sourced from a [GitHub Secret][gh-secret]. If unspecified, other
+    authentication methods are attempted.
+
+
+## Outputs
```
craigdbarber: Suggest adding an example.

get-secretmanager-secrets/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "get-secretmanager-secrets",
+  "version": "0.1.0",
+  "description": "Get Secret Manager secrets GitHub action",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "ncc build src/main.ts",
+    "test": "mocha -r ts-node/register 'tests/*.spec.ts'"
```
craigdbarber: Please include scripts for format and format-check using prettier, see: https://github.com/GoogleCloudPlatform/github-actions/blob/9ef0382a75df2efcc72850f4e963927a45d59414/setup-gcloud/package.json#L11
Also, please include a precommit script: https://github.com/GoogleCloudPlatform/github-actions/blob/9ef0382a75df2efcc72850f4e963927a45d59414/setup-gcloud/package.json#L12

get-secretmanager-secrets/src/client.ts
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+import { GoogleAuth } from 'google-auth-library';
+
+type ClientOptions = {
+  credentials?: string,
+  endpoint?: string,
+}
+
+export class Client {
```
craigdbarber: Please be sure to document classes and methods with TypeDoc for consistency with the rest of the repo. Here and throughout.

.github/workflows/get-secretmanager-secrets-it.yml
sethvargo: It's masked 😄but the previous action. Also, it's not a _real_ secret, so I'm not concerned about leaking it.

get-secretmanager-secrets/src/client.ts
```diff
@@ -0,0 +1,81 @@
...
+  readonly defaultEndpoint: string = 'https://secretmanager.googleapis.com/v1beta1';
+  readonly defaultScope: string = 'https://www.googleapis.com/auth/cloud-platform';
+  readonly userAgent = 'github-action-google-secretmanager/0.1.0';
+
+  readonly auth: GoogleAuth;
+  readonly endpoint: string;
+
+  constructor(opts?: ClientOptions) {
+    this.endpoint = opts?.endpoint || this.defaultEndpoint;
```
craigdbarber: Should add a check to ensure opts is defined otherwise could get a runtime error.

get-secretmanager-secrets/src/client.ts
```diff
@@ -0,0 +1,81 @@
...
+    } else {
+      this.auth = new GoogleAuth({
+        scopes: [this.defaultScope],
+        projectId: 'unused-but-required',
+      });
+    }
+  }
+
+  async accessSecret(ref: string): Promise<string> {
+    const client = await this.auth.getClient();
```
craigdbarber: Should add a precondition check ensuring ref is not empty.

get-secretmanager-secrets/src/client.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    const headers = await client.getRequestHeaders();
+    headers['User-Agent'] = this.userAgent;
+
+    const url = `${this.endpoint}/${ref}:access`;
```
craigdbarber: Why are we doing raw REST calls instead of using the KMS client lib?

get-secretmanager-secrets/src/client.ts
```diff
@@ -0,0 +1,81 @@
...
+    const resp = await client.request({
+      url: url,
+      headers: headers,
+    }) as PayloadResponse;
+    const b64data = resp.data.payload.data;
```
craigdbarber: Should add a checks to ensure these child objects are defined.

get-secretmanager-secrets/src/reference.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (sParts.length < 2) {
+      throw new TypeError(`Invalid reference "${s}" - missing destination`);
+    }
+
+    this.output = sParts[0].trim();
+
+    const ref = sParts.slice(1).join(":");
+    const refParts = ref.split("/");
+    switch (refParts.length) {
+      case 6: {
```
craigdbarber: Could you please add a comment for each case showing an example of the expected format.

get-secretmanager-secrets/tests/client.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { expect } from "chai";
+import "mocha";
+
+import { Client } from "../src/client";
+
+const secretRef = process.env.GITHUB_ACTIONS_GET_SECRETMANAGER_SECRET_SECRET_REF!;
+
+describe("Client", () => {
+  it("initializes with JSON creds", () => {
```
craigdbarber: Please also add tests for failure cases.

get-secretmanager-secrets/tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+    "lib": [
+      "es6"
+    ],
+    "outDir": "dist",
+    "rootDir": "./src",
+    "strict": true,
+    "noImplicitAny": true,
+    "esModuleInterop": true
+  },
+  "exclude": ["node_modules", "tests"]
```
craigdbarber: Please use the same exclude pattern for tests as setup-gcloud: https://github.com/GoogleCloudPlatform/github-actions/blob/66f24ece9d2631f3ce0f06dddf4df0febe9a7eaa/setup-gcloud/tsconfig.json#L29

get-secretmanager-secrets/src/client.ts
```diff
@@ -0,0 +1,81 @@
...
+
+type ClientOptions = {
+  credentials?: string,
+  endpoint?: string,
+}
+
+export class Client {
+  readonly defaultEndpoint: string = 'https://secretmanager.googleapis.com/v1beta1';
+  readonly defaultScope: string = 'https://www.googleapis.com/auth/cloud-platform';
+  readonly userAgent = 'github-action-google-secretmanager/0.1.0';
```
craigdbarber: Let's keep the userAgent format consistent with the setup-gcloud action: https://github.com/GoogleCloudPlatform/github-actions/blob/66f24ece9d2631f3ce0f06dddf4df0febe9a7eaa/setup-gcloud/src/install-util.ts#L26

get-secretmanager-secrets/package.json
sethvargo: 1. What is the value in having `format` and `format-check` when the `eslint` command automatically runs prettifier and fixes any issues?

1. Adding a precommit hook like that doesn't actually work and isn't working in setup-gcloud either. Husky requires a new top-level config and doesn't live under `scripts` anymore ([ref](https://github.com/typicode/husky#upgrading-from-014)). I'm happy to fix it, but can you help me understand what problem it's solving?

get-secretmanager-secrets/src/client.ts
sethvargo: Typescript handles this as of 3.7: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining (we are using 3.7.5). For example, this code compiles to:

```javascript
this.endpoint = ((_a = opts) === null || _a === void 0 ? void 0 : _a.endpoint) || this.defaultEndpoint;
```

which does the proper null check.

get-secretmanager-secrets/src/client.ts
sethvargo: 1. Because this is Secret Manager, not KMS

1. Importing Secret Manager imports all the GAPIC and GRPC libraries which are super heavy for this type of work. We only need to make a single API call, and it's not worth the 14.6x increase in the resulting package size.

.github/workflows/get-secretmanager-secrets-it.yml
sethvargo: Done in https://github.com/GoogleCloudPlatform/github-actions/pull/53/commits/b060325aeba4554d6e9983f16f84411c8c838e13. I took a slightly different approach with using a boolean field instead. There's no value in having a user supply a filepath because that volume needs to be shared among all builds steps for the ADC to actually work. Instead, I write it to the shared workspace. 

Default behavior is off, so it's fully BC. I also documented it, wrote tests, and updated the action specification.

sethvargo:
Hey @craigdbarber - thanks. I think I addressed all feedback sans one item about some script helpers and husky.

get-secretmanager-secrets/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "get-secretmanager-secrets",
+  "version": "0.1.0",
+  "description": "Get Secret Manager secrets GitHub action",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "ncc build src/main.ts",
+    "lint": "eslint . --ext .ts,.tsx",
```
craigdbarber: Please use prettier for this step to ensure formatting consistency with the rest of the repo.

get-secretmanager-secrets/src/client.ts
craigdbarber: TIL, thanks for clarifying :)

get-secretmanager-secrets/src/client.ts
craigdbarber: Makes sense. This might be good feedback to pass along to the DPE team that maintains the libraries. Separately, is there any concern about the underlying REST APIs changing?

get-secretmanager-secrets/src/client.ts
```diff
@@ -0,0 +1,113 @@
...
+    } else {
+      this.auth = new GoogleAuth({
+        scopes: [this.defaultScope],
+        projectId: 'unused-but-required',
+      });
+    }
+  }
+
+  /**
+   * accessSecret retrieves the secret by name.
```
craigdbarber: No need to repeat the name of the method in the header, here and throughout.

get-secretmanager-secrets/tests/client.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    });
+    expect(client.auth.jsonContent).eql({ foo: 'bar' });
+  });
+
+  it('initializes with ADC', () => {
+    const client = new Client();
+    expect(client.auth.jsonContent).eql(null);
+  });
+
+  it('accesses secret versions', async function() {
```
craigdbarber: s/function()/() =>

get-secretmanager-secrets/package.json
sethvargo: It does - this is the recommended settings from prettier. The eslintrc extends `plugin:prettier/recommended`, which applies the prettier settings.

get-secretmanager-secrets/tests/client.test.ts
sethvargo: That won't work as arrow bindings dont work with `this` (which is needed to call `skip()`)

get-secretmanager-secrets/package.json
sethvargo: https://prettier.io/docs/en/integrating-with-linters.html#eslint

get-secretmanager-secrets/package.json
sethvargo: I added husky with the new config. Still interested in your feedback on 1.

get-secretmanager-secrets/package.json
craigdbarber: Makes sense. In that case, would you mind updating the settings in the setup-gcloud action to be consistent?

get-secretmanager-secrets/tests/client.test.ts
craigdbarber: Makes sense. Thanks for clarifying.

get-secretmanager-secrets/package.json
sethvargo: I think I'd prefer to do that in a followup PR. The diff is pretty large: https://github.com/sethvargo/github-actions/pull/2

PR #60, opened 2020-02-21
Update linter and formatter to match

sethvargo:
Match linter and esrules

Do you want me to also update husky in this PR?


craigdbarber:
> Match linter and esrules
> 
> Do you want me to also update husky in this PR?

Sure, that'd be great. Thanks!

setup-gcloud/tests/download-util.test.ts
```diff
@@ -0,0 +1,89 @@
...
+describe('#downloadAndExtractTool', function() {
+  beforeEach(async function() {
+    await io.rmRF(toolDir);
+    await io.rmRF(tempDir);
+  });
+
+  afterEach(async function() {
+    await io.rmRF(toolDir);
+    await io.rmRF(tempDir);
+  });
```
craigdbarber: Please be sure to preserve the original test timeouts, here and throughout. The MacOS runners can be very flaky if you don't

craigdbarber:
Thanks for the contribution! One general note, using git mv is preferred over git rm and git add, as it preserves the history for the file.

setup-gcloud/tests/download-util.test.ts
sethvargo: Each test has a an individual timeout of 30s per the mocha command in the package.json. Do you want it to be higher?

sethvargo:
@craigdbarber ack, sorry

setup-gcloud/tests/download-util.test.ts
craigdbarber: 30s works. Thanks for clarifying :)

Issue #65, opened 2020-03-06
UUID file left behind after run with export_default_credentials=true


enochcheung:
With export_default_credentials=true, after the setup-gcloud Github Action is run, a file is left behind in the root of the repository. The filename looks like just a GUID. This leaves the repository in an unclean state that can mess with some builds.

Seems that this is a file used to persist credentials to downstream steps
https://github.com/GoogleCloudPlatform/github-actions/blob/5f5ef3df93d08dce64f649240ddd69b89cab7f19/setup-gcloud/src/setup-gcloud.ts#L88

I don't know if this is avoidable, but since this is pretty surprising behavior it would be nice if it was documented

sethvargo:
Hi @enochcheung - thanks for opening an issue. That file has to persist (it's how the credentials are "exported"). Do you have ideas about ways we could better document it?

enochcheung:
I think you could just add it to the readme? Opened this PR https://github.com/GoogleCloudPlatform/github-actions/pull/66 but feel free close it and word it differently etc

sethvargo:
Thanks!

sethvargo:
Documented in #66

PR #66, opened 2020-03-06
Document export_default_credentials file creation

enochcheung:
Document the fact that turning on export_default_credentials will cause a file to be generated

Addresses https://github.com/GoogleCloudPlatform/github-actions/issues/65


enochcheung:
@googlebot I signed it!

PR #68, opened 2020-03-10
Allow setting a default project ID

### LINKED ISSUE #63: gsutil requires a project id
When running `gsutil` in an action.yml file, an error is thrown requiring a project id during configuration.

```
    - run: |
        # View storage
        gsutil ls

You are attempting to perform an operation that requires a project id, with none configured. Please re-run gsutil config and make sure to follow the instructions for finding and entering your default project id.
```

However if *gcloud* is activated with a valid *project_id*, then, the gsutil command works correctly.  For example, like the following:

```
gcloud auth activate-service-account --project=${someProject_id} --key-file=${tmpKeyFilePath}`
```

The problem may be here https://github.com/GoogleCloudPlatform/github-actions/blob/2a9cbc618184da4718e0d16a1b0a055a4f90f66e/setup-gcloud/src/setup-gcloud.ts#L84

Note there is no **--project**, which could be picked up from the key-file.

sethvargo:
This adds a new configuration option, `project_id`. If provided, the runner executes `gcloud config set core/project <value>`, which can save time. Individual actions can still override project using `--project` and this is opt-in behavior so no existing actions will be impacted.

Fixes #63


sethvargo:
Oh, this is based on https://github.com/GoogleCloudPlatform/github-actions/pull/67, so once that's reviewed/merged, I can rebase this

sethvargo:
This requires a new secret `SETUP_GCLOUD_IT_PROJECT_ID` to be created too please ❤️

setup-gcloud/src/setup-gcloud.ts
```diff
@@ -87,17 +108,15 @@ async function run(): Promise<void> {
...
-    // A workaround for https://github.com/actions/toolkit/issues/229
-    // Currently exec on windows runs as cmd shell.
-    let toolCommand = 'gcloud';
-    if (process.platform == 'win32') {
-      toolCommand = 'gcloud.cmd';
-    }
-
-    // authenticate as the specified service account
-    await exec.exec(
-      `${toolCommand} auth activate-service-account ${serviceAccountEmail} --key-file=${tmpKeyFilePath}`,
```
sethvargo: So, fun fact (for some definition of "fun"), we weren't escaping any of these values, so you can RCE yourself 😄

.github/workflows/setup-gcloud-it.yml
```diff
@@ -69,3 +69,22 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        operating-system: [ubuntu-latest, windows-latest, macos-latest]
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: setup-gcloud
+        uses: ./setup-gcloud/
+        with:
+          project_id: ${{ secrets.SETUP_GCLOUD_IT_PROJECT_ID }}
```
craigdbarber: Still need to specify the GCP SA key to use.

setup-gcloud/src/setup-gcloud.ts
craigdbarber: Good catch!

.github/workflows/setup-gcloud-it.yml
sethvargo: The project setting happens before authentication, so you don't actually need the service account here. It just skips service account activation if one is not given. We do need to create the `SETUP_GCLOUD_IT_PROJECT_ID` secret though to get the tests to pass.

craigdbarber: Ack, thanks for clarifying.

sethvargo:
Done - pushed to sethvargo/project

sethvargo:
Looks like the secret is not created? https://github.com/GoogleCloudPlatform/github-actions/actions/runs/53798008

craigdbarber:
Secret added, re-running jobs.

craigdbarber:
Looks like the macos runner is timing out in the unit tests. Mind increasing the timeout in this PR to assuage my OCD need for green? :)

sethvargo:
All happy: https://github.com/GoogleCloudPlatform/github-actions/actions/runs/54546388

craigdbarber:
Awesome! Tyvm :)

PR #70, opened 2020-03-13
[Action] Deploy to App Engine

averikitsch:
I've pulled all the gcloud setup into a separate package which is used by the action. After this PR I will refactor the setup-gcloud actions to use the new package.

I need help know how to proceed forward with have the separate package. Are we going to publish to npm or how can we refer to it in typescript? 
Also, I would like to make sure there are enough tests.


appengine-deploy/src/main.ts
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+import * as core from '@actions/core';
+import * as exec from '@actions/exec';
+import * as setupGcloud from '../../pkg/dist/index';
+
+async function run(): Promise<void> {
+  try {
+    // Get action inputs.
+    const projectId = core.getInput('project_id', { required: true });
```
sethvargo: Does this need to be required? If the user set the project via gcloud, then this wouldn't be required.

appengine-deploy/src/main.ts
```diff
@@ -0,0 +1,101 @@
...
+    const deliverables = core.getInput('deliverables');
+    const imageUrl = core.getInput('image-url');
+    const version = core.getInput('version');
+    const promote = core.getInput('promote');
+
+    if (!setupGcloud.isInstalled()) {
```
sethvargo: My initial thought is two fold:

1. If we're going to use gcloud, we should clearly document the requirement on the `setup-gcloud` action and throw an error here if it's not installed instead of duplicating the logic.

2. We shouldn't depend on gcloud here unless we absolutely have to. It looks like we can create an AppEngine deployment using the API directly: https://cloud.google.com/appengine/docs/admin-api/deploying-overview

appengine-deploy/tsconfig.json
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright 2019 Google LLC
```
sethvargo: 20

pkg/dist/format-url.d.ts
```diff
@@ -0,0 +1,10 @@
+/**
```
sethvargo: Dist should only include the compiled index.js file. I think these belong in src?

appengine-deploy/src/main.ts
averikitsch: This is for the case that gcloud is authenticated with one project/service account but wanting to deploy to another project (with the correct permissions).

appengine-deploy/src/main.ts
averikitsch: 1. There isn't a dependency on `setup-gcloud`. It can be used with or without, so I check if gcloud is installed. 
2. I debated `gcloud` vs using the API directly. Using the API directly means reimplementing gcloud functionality, such as: Uploading files to GCS, Authentication, yaml parsing, Creating and managing the GAE application services and versions, Java specifics (deploying appengine-web.xml, jars), and gcloud flags (promote/no promote, flex container images, version naming, multiple deployables).

appengine-deploy/src/main.ts
sethvargo: Right, but I don't think it should be _required_. It could be optional right? I think we should remove the `required: true` attribute here.

appengine-deploy/src/main.ts
sethvargo: Sure, but you're doing a _lot_ of duplication of setup-gcloud here, including downloading and installing. It seems like we should remove all of that and require the user has the `setup-gcloud` action in their pipeline. We can just fail if you don't find the `gcloud` CLI command.

appengine-deploy/src/main.ts
averikitsch: Got it, I see what you mean now.

appengine-deploy/src/main.ts
averikitsch: Right, I've created a separate package to host the shared code between setup-gcloud and this action. After I was going to refactor the setup-gcloud method. I didn't want the action to require the setup-gcloud action. I am definitely open to switching it over if you think that is the best.

sethvargo: It's my _opinion_ that setup-gcloud should be the only thing that installs/configures gcloud. If other actions require the `gcloud` CLI or authentication, they should document that requirement, but not duplicate the installation. I think it would be surprising to users to suddenly see a new tool in their path.

This also matches the way other actions work. For example, if your action requires python, you need to first use the `setup-python` action. Those actions don't magically install python because that would be unexpected.

appengine-deploy/README.md
```diff
@@ -0,0 +1,129 @@
...
+    project_id: ${{ secrets.PROJECT_ID }}
+
+# Example of using the output
+- id: test
+  run: curl "${{ steps.deploy.outputs.url }}"
+```
+
+## Inputs
+
+- `project_id`: (Required)
```
craigdbarber, line 43: When possible let's try to keep consistency with other actions documentation for common fields like this. See: https://github.com/GoogleCloudPlatform/github-actions/tree/master/setup-gcloud

appengine-deploy/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+- `deliverables`: The yaml files for the services or configurations you want to
```
craigdbarber, line 45: Consider adding a link to the documentation for the App Engine config file format.

appengine-deploy/README.md
```diff
@@ -0,0 +1,129 @@
...
+```yaml
+- uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+  with:
+    export_default_credentials: true
+- uses: GoogleCloudPlatform/github-actions/get-secretmanager-secrets@master
+```
+
+### Via Credentials
+
+You can provide [Google Cloud Service Account JSON][sa] directly to the action
```
craigdbarber, line 96: s/sa/SA

appengine-deploy/src/main.ts
```diff
@@ -0,0 +1,101 @@
...
+      appDeployCmd.push('--version', version);
+    }
+    if (promote) {
+      appDeployCmd.push('--promote');
+    } else {
+      appDeployCmd.push('--no-promote');
+    }
+
+    // Get output of gcloud cmd.
+    let myOutput = '';
```
craigdbarber, line 75: Please go throw and mitigate these GitHub warnings throughout.

appengine-deploy/src/main.ts
```diff
@@ -0,0 +1,101 @@
...
+          myError += data.toString();
+        }
+      },
+    };
+
+    // Run gcloud cmd.
+    await exec.exec(toolCommand, appDeployCmd, options);
+
+    // Set url as output.
+    const url = myError.match(
```
craigdbarber, line 93: On a successful deploy, shouldn't the url be getting pulled from myOutput?

craigdbarber:
Great contribution! Regarding the library question, since this is the first consumer of the lib I think we can just use the internal code reference for now, and circle back to the discussion of publishing an npm package as needed.

appengine-deploy/tsconfig.json
averikitsch: done

pkg/dist/format-url.d.ts
averikitsch: In order to use the package I have to set the compiler field composite to true which then automatically adds these to the dist.

appengine-deploy/README.md
averikitsch: done

appengine-deploy/README.md
averikitsch: Does this need to be changed based on styling?

.github/workflows/appengine-deploy-it.yml
```diff
@@ -0,0 +1,44 @@
...
+on: [push, pull_request]
+jobs:
+  gcloud:
+    name: with setup-gcloud
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: ./setup-gcloud
+      with:
+        service_account_email: ${{ secrets.APPENGINE_DEPLOY_SA_EMAIL }}
```
averikitsch, line 11: @craigdbarber Can you help me set up secrets for this repo? Do we have a test project?

craigdbarber: @averikitsch happy to :)
Will sync offline.

appengine-deploy/README.md
craigdbarber: My bad, didn't notice this was a link. Disregard :)

pkg/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+import { promises as fs } from 'fs';
+import { getReleaseURL } from './format-url';
+import * as downloadUtil from './download-util';
+import * as installUtil from './install-util';
+import { getLatestGcloudSDKVersion } from './version-util';
+
+export { getLatestGcloudSDKVersion };
+
+/**
+ * @Method: Check if gcloud is installed
```
craigdbarber, line 30: Let's keep the documentation consistent across the repo with TypeDoc, see: https://typedoc.org/guides/doccomments

pkg/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "pkg",
+  "version": "1.0.0",
+  "description": "Common tools needed to install and use the Cloud SDK",
+  "main": "dist/index.js",
```
craigdbarber, line 5: Shouldn't this also include a "types" element?

pkg/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "pkg",
```
craigdbarber, line 2: Suggest using a library name that's a bit more descriptive. Perhaps setupGloudSDK?

pkg/dist/download-util.d.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Downloads and extracts the tool at the specified URL.
+ *
+ * @url The URL of the tool to be downloaded.
+ * @returns The path to the locally extracted tool.
+ */
+export declare function downloadAndExtractTool(url: string): Promise<string>;
```
craigdbarber, line 7: Shouldn't the library's index.d.ts contain all exported types for the lib? I'm not sure why each individual sub-component needs a template file in dist.

craigdbarber:
Looks great. Would love to see the re-factor work for the setupGcloud action also included :)

pkg/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+ * @Param {string} version (Optional) Cloud SDK version
+ * @Return {string} tool path for Cloud SDK
+ */
+export function isInstalled(version?: string) {
+  let toolPath;
+  if (version) {
+    toolPath = toolCache.find('gcloud', version);
+  } else {
+    toolPath = toolCache.findAllVersions('gcloud');
```
sethvargo, line 39: Note, this could return an array if there are multiple gcloud versions installed.

averikitsch:
Closing this PR to push to a non-forked branch.

PR #77, opened 2020-03-17
Add GitHub issue template

sethvargo:
This adds an initial issue template (we can iterate on it in the future) to help reduce the back-and-forth with bug reports.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,33 @@
+---
+name: Bug report
+about: Tell us about a bug.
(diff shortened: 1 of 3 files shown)


bharathkkb:
@sethvargo can we have three templates for bug/feature/question like [we have here]( https://github.com/terraform-providers/terraform-provider-google/tree/master/.github/ISSUE_TEMPLATE)?
This will give folks an option when opening an issue like this:
image

sethvargo:
Done

sethvargo:
Gonna merge this for now since we're getting more usage. We can always tweak these as we go along 😄

PR #78, opened 2020-03-17
Only run CI checks that correspond to changed files

sethvargo:
This reduces the number of tests we need to run by mapping the tests to the files changed.


sethvargo:
Hey @craigdbarber what do you think about this? I think it's going to make iterating on a single action a lot easier and also reduce the number of builds we do per PR>

sethvargo:
Craig is OOO - @averikitsch could you PTAL and let me know what you think?

Issue #84, opened 2020-03-30
[setup-gcloud] already downloaded gcloud binary not found on second and subsequent run


jacek-jablonski:
### TL;DR
I am running self hosted runner and Google Cloud SDK is NOT installed on system. At first run setup-gcloud downloads GoogleCloud SDK and adds gcloud binary location to PATH. Everything works fine. On second run, gcloud cache is found, but gcloud location is not added to PATH. It results in error.

On Github's runner it runs fine, probably because Google Cloud SDK is installed on their VMs system-wide.

**Expected behavior**
Already cached gcloud is added to PATH and action runs fine.

**Observed behavior**
```
::debug::isExplicit: 286.0.0
::debug::explicit? true
::debug::checking cache: /home/actions/cache/gcloud/286.0.0/x64
::debug::Found tool in cache gcloud 286.0.0 x64
::error::Unable to locate executable file: gcloud. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.
```

### Reproduction

**Action YAML**

```yaml
# Paste your complete GitHub Actions YAML here, removing
# any sensitive values.
```

**Repository**

**Additional information**
Running on self-hosted runner.

sethvargo:
Hey @jacek-jablonski - thank you for opening a detailed issue. Can you provide more information about your custom runner setup? If you have some Terraform or gcloud scripts that you used to create them, that would be super helpful for us to reproduce and fix the issue.

jacek-jablonski:
Hi @sethvargo,
Below you can find Dockerfile I'm using:
```
FROM ubuntu:bionic

ARG GIT_VERSION="2.23.0"
ARG GH_RUNNER_VERSION="2.165.2"
ARG TARGETPLATFORM="x64"

SHELL ["/bin/bash", "-o", "pipefail", "-c"]
# hadolint ignore=DL3003
RUN apt-get update && \
  apt-get install -y --no-install-recommends \
    apt-transport-https \
    ca-certificates \
    gnupg-agent \
    curl \
    git \
    software-properties-common \
    python-openssl \
  && rm -rf /var/lib/apt/lists/*

RUN curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add - \
[...]

PR #85, opened 2020-03-30
fix cached gcloud path

jacek-jablonski:
fixes: https://github.com/GoogleCloudPlatform/github-actions/issues/84


jacek-jablonski:
@googlebot I signed it!

sethvargo:
Ah I see. So we have an edge case where, if the tool is already installed in the cache, we don't add it to the PATH? @craigdbarber what do you think about this change?

PR #90, opened 2020-04-10
Update kubectl rollout status command

anthonymonori:
In the current example it is hardcoded to `gke-test` but that will fail if people follow the tutorial in the README.md and change the IMAGE name


anthonymonori:
@googlebot I signed it!

sethvargo:
Thanks!

Issue #95, opened 2020-04-28
setup-gcloud: authentication fields description is misleading


IlyaSemenov:
### TL;DR
The documentation for setup-gcloud description is misleading:

> If `service_account_key` is specified, authenticates the gcloud CLI tool
    using the inputs: `service_account_email` and `service_account_key`. Please
    see the [Service Account documentation][sa-iam-docs] for more information.

`service_account_email` is listed first, however in fact only `service_account_key` is used, as email is pulled from the key (#8, #9).

**Expected behavior**

The documentation must clearly reflect that only `service_account_key` is needed for authentication, and also outline the specific circumstances when `service_account_email` should be used.

sethvargo:
`service_account_email` is [used if given](https://github.com/GoogleCloudPlatform/github-actions/blob/20c294aabd5331f9f7b8a26e6075d41c31ce5e0d/setup-gcloud/src/setup-gcloud.ts#L118), but it defaults to the empty string because it can be inferred from _certain types_ of keys, specifically, the value is required when using the older format `.p12` keys.

Providing the service account email for JSON verifies the email matches the one in the key, which can be used as an extra sanity check.

IlyaSemenov:
I am not sure which reply do you expect? Sure thing, I understand that service account email might be useful in certain circumstances, including the older key format.

It is however not needed in most *mainstream* operations:
* It is not used by the [sister GitHub action](https://github.com/actions-hub/gcloud).
* It is not needed in other similar flows such as [Google Cloud Getting Started](https://cloud.google.com/docs/authentication/getting-started#setting_the_environment_variable).

so the current wording where `service_account_email` is unquestionably placed first is a bit misleading to my thinking. If you don't concur to that, just close the ticket.

(That was a friendly feedback from someone who just realised that one of his CI/CD secrets doesn't serve any purpose - other than the sanity check, indeed - and was happy to simplify the setup, nothing more.)

sethvargo:
Thanks, opened https://github.com/GoogleCloudPlatform/github-actions/pull/96

PR #96, opened 2020-04-28
Add clarity around service_account_email

sethvargo:
Fixes GH-95

### DIFF
--- a/.github/workflows/get-secretmanager-secrets-it.yml
+++ b/.github/workflows/get-secretmanager-secrets-it.yml
@@ -13,7 +13,6 @@ jobs:
     - uses: actions/checkout@v2
     - uses: ./setup-gcloud
       with:
-        service_account_email: ${{ secrets.GET_SECRETMANAGER_SECRETS_SA_EMAIL }}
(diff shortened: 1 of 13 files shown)


bharathkkb:
@sethvargo something I just noticed, we require [service_account_key](https://github.com/GoogleCloudPlatform/github-actions/blob/20c294aabd5331f9f7b8a26e6075d41c31ce5e0d/setup-gcloud/action.yml#L41) but mark it as [optional in the Readme](https://github.com/GoogleCloudPlatform/github-actions/blob/master/setup-gcloud/README.md#inputs)

sethvargo:
@bharathkkb it's only required to setup auth. If you don't want to configure auth (i.e. you just want gcloud), or if you want to export just the project ID, you don't need to provide a key.

bharathkkb:
@sethvargo yup makes sense and just the functionality I was looking for 
```
Successfully installed gcloud Cloud SDK
No credentials provided, skipping authentication
```
I assumed it would fail because it was marked required in the action.yml

sethvargo:
Oh sorry, I misunderstood your comment. Updated.

PR #115, opened 2020-05-30
Set project ID from service account

### LINKED ISSUE #114: setup-gcloud - Automatically set project from service account if not specified
### TL;DR
If the user sets a service_account, I think this demonstrates clear intent to use it with a project. If they don't set the project explicitly, we leave it unset. Since all service accounts must be associated with a project, would be convenient, and almost expected, to set the project as the project that the service account belongs to.

### Design
If project_id is not set, and service account is set, set the project to the service account's project.

Yaml would have one less line required.

Mark-McCracken:
Resolves #114

### DIFF
--- a/setup-gcloud/README.md
+++ b/setup-gcloud/README.md
@@ -26,8 +26,10 @@ It does the following:
 1.  Installs and caches the downloaded version into the actions environment.
 
 1.  If `project_id` is specified, gcloud will use this project ID as the default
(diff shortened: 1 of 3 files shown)


sethvargo:
Thanks for the PR. Left my feedback on #114.

sethvargo:
Closing per discussion on #114.

PR #117, opened 2020-06-03
Cloud Run Action

averikitsch:


deploy-cloudrun/action.yml
```diff
@@ -0,0 +1,71 @@
...
+    description: |-
+      ID of the service or fully qualified identifier for the service.
+      Required if not using a service YAML.
+    required: false
+
+  region:
+    description: |-
+      Region in which the resource can be found.
+    required: false
+    default: us-central1
```
craigdbarber, line 39: Is this a conventional default region with other GCP services? Just want to make sure we're following established conventions vs defining our own.

deploy-cloudrun/src/cloudRun.ts
```diff
@@ -0,0 +1,258 @@
...
+        creds = Buffer.from(creds, 'base64').toString('utf8');
+      }
+      const jsonContent = JSON.parse(creds);
+      this.auth.jsonContent = jsonContent;
+
+      if (!projectId) {
+        projectId = jsonContent.project_id;
+        core.info('Setting project Id from credentials');
+      }
+    } else if (!projectId) {
```
craigdbarber, line 91: It might be clearer if the assumptions are stated explicitly within the conditional statement as opposed to transitively like this. Consider something like:
`if (!projectId && json.project_id) {}
elseif (!projectId && process.env.GCLOUD_PROJECT) {}
else {throw new Error()}`

deploy-cloudrun/src/cloudRun.ts
```diff
@@ -0,0 +1,258 @@
...
+ * @param region Region of the GCP resource.
+ * @param opts list of ClientOptions
+ * @returns CloudRun client
+ */
+export class CloudRun {
+  readonly methodOptions = {
+    userAgentDirectives: [
+      {
+        product: 'github-actions-deploy-cloudrun',
+        version: '0.1.0',
```
craigdbarber, line 51: I wonder if there's a way to pull this from the env. That way we don't have to remember to update this string with every release.

deploy-cloudrun/src/index.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    // Create Cloud Run client
+    const client = new CloudRun(region, { projectId, credentials });
+
+    // Initialize service
+    const service = new Service({ image, name, envVars, yaml });
+
+    // Deploy service
+    let serviceResponse = await client.deploy(service);
+    while (!serviceResponse.status!.url) {
```
craigdbarber, line 44: What does the error messaging look like to the user with these non-null assertions? I want to be mindful of the potential errors messages which might be seen by users. I wonder if it make sense to do explicit null checks here and throughout so that we can control the error messaging.

deploy-cloudrun/src/service.ts
```diff
@@ -0,0 +1,132 @@
...
+    this.name = request.metadata!.name!;
+  }
+
+  /**
+   * Parses a string of the format `KEY1=VALUE1`.
+   *
+   * @param envVarInput Env var string to parse
+   * @returns EnvVar[]
+   */
+  protected parseEnvVars(envVarInput: string): EnvVar[] {
```
craigdbarber, line 119: Not necessarily in this PR, but this seems like something which could be useful across multiple actions. Could be worth pulling out into something more centralized.

deploy-cloudrun/src/cloudRun.ts
```diff
@@ -0,0 +1,258 @@
...
+    core.info(`Service ${service.name} has been successfully deployed.`);
+    return serviceResponse.data;
+  }
+
+  /**
+   * Deletes a Cloud Run service.
+   *
+   * @param service Service object
+   */
+  async delete(service: Service): Promise<void> {
```
bharathkkb, line 213: Are we supporting a delete feature or is this just for cleaning up the test?
https://github.com/GoogleCloudPlatform/github-actions/blob/828bf4e608a7cc05a79e539b6bb9b5354bdce7eb/deploy-cloudrun/tests/unit/cloudRun.test.ts#L61

deploy-cloudrun/src/cloudRun.ts
```diff
@@ -0,0 +1,258 @@
...
+      core.info(`Error deleting Service ${service.name}: ` + e);
+    }
+  }
+
+  /**
+   * Set's IAM policy for service (Not Recommended)
+   *
+   * @param service Service object
+   */
+  async allowUnauthenticatedRequests(service: Service): Promise<void> {
```
bharathkkb, line 235: Are we supporting this?

deploy-cloudrun/README.md
```diff
@@ -0,0 +1,178 @@
...
+limitations under the License.
+-->
+# deploy-cloudrun
+
+This action deploys your container image to [Cloud Run][cloud-run] and makes the URL
+available to later build steps via outputs.
+
+## Prerequisites
+
+This action requires Google Cloud credentials that are authorized to deploy a
```
bharathkkb, line 23: Shouldn't we mention that `run.googleapis.com` need to be enabled?

deploy-cloudrun/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+- See [Deploy a new revision of an existing service](https://cloud.google.com/run/docs/deploying#yaml_1)
+to generated a YAML service specification from an existing service:
+
+```
+gcloud run services describe SERVICE --format yaml > service.yaml
+```
+## Allow unauthenticated requests
+
+A Cloud Run product recommendation is that CI/CD systems not set the
```
bharathkkb, line 98: I think this wording does not make sense with respect to the run action as `--allow-unauthenticated` is a gcloud flag and we are not exposing any option for this in the action.

deploy-cloudrun/action.yml
averikitsch: Yes per DevRel but not officially, I can remove and make this required?

deploy-cloudrun/src/cloudRun.ts
averikitsch: Just for removing test resources

deploy-cloudrun/src/cloudRun.ts
averikitsch: Not supporting, but I was keeping for prosperity. Should I remove?

deploy-cloudrun/README.md
averikitsch: Added.

deploy-cloudrun/README.md
averikitsch: updated.

deploy-cloudrun/src/cloudRun.ts
averikitsch: Submitted a question / feature request https://github.com/actions/toolkit/issues/494

deploy-cloudrun/src/index.ts
averikitsch: If there is a `serviceResponse` obj then there will be a `status` field. If that breaks then it most likely with error: `TypeError: Cannot read property 'url' of undefined`.

Issue #122, opened 2020-06-06
AppEngine deploy tests cannot run in parallel


sethvargo:
### TL;DR
Conflicts due to same name being deployed.

**Expected behavior**
Tests pass

**Observed behavior**
Tests fail ([example](https://github.com/GoogleCloudPlatform/github-actions/pull/118/checks?check_run_id=743985273))

```
File upload done.
ERROR: (gcloud.app.deploy) ABORTED: Cannot operate on apps/***/services/default/versions/gcloud because an operation is already in progress for apps/***/services/default/versions/gcloud by 04ecd77a-b346-4d7f-bdbc-25c0b8139baf.
##[error]The process '/opt/hostedtoolcache/gcloud/295.0.0/x64/bin/gcloud' failed with exit code 1
```

PR #123, opened 2020-06-06
Allow GAE integration tests to run in parallel

sethvargo:
Fixes GH-122


.github/workflows/appengine-deploy-it.yml
```diff
@@ -10,52 +15,66 @@ jobs:
...
       name: Deploy to App Engine
       uses: ./appengine-deploy
       with:
         project_id: ${{ secrets.APPENGINE_DEPLOY_PROJECT_ID }}
-        deliverables: ./appengine-deploy/test/app.yaml
+        deliverables: ${{ github.workspace }}/app.yaml
         version: gcloud
     - name: Test Output
-      run: curl '${{ steps.deploy.outputs.url }}'
+      run: curl -sfL '${{ steps.deploy.outputs.url }}'
```
sethvargo, line 32: Curl doesn't actually exit non-zero by default

.github/workflows/appengine-deploy-it.yml
```diff
@@ -1,5 +1,10 @@
 name: appengine-deploy Integration
-on: [push, pull_request]
+
+on:
+  push:
+    paths:
+    - 'appengine-deploy/**'
```
sethvargo, line 6: Secret's are available in Pull Requests, so only run on push for now. These will still run when _we_ push, since it's not restricted to the master branch.

.github/workflows/appengine-deploy-it.yml
```diff
@@ -10,52 +15,66 @@ jobs:
       with:
         service_account_key: ${{ secrets.APPENGINE_DEPLOY_SA_KEY_JSON }}
         export_default_credentials: true
+    - name: create yaml
+      run: |-
+        cat <<EOF > ${{ github.workspace }}/app.yaml
+        service: "${{ github.job }}-${{ github.run_number }}"
```
sethvargo, line 21: This should be "unique enough"

sethvargo:
@averikitsch

.github/workflows/appengine-deploy-it.yml
```diff
@@ -10,52 +15,66 @@ jobs:
...
+        EOF
     - id: deploy
       name: Deploy to App Engine
       uses: ./appengine-deploy
       with:
         credentials: ${{ secrets.APPENGINE_DEPLOY_SA_KEY_JSON }}
-        deliverables: ./appengine-deploy/test/app.yaml
-        version: json
+        deliverables: ${{ github.workspace }}/app.yaml
+        version: ${{ github.job }}-${{ github.run_number }}-json
```
averikitsch, line 75: Is this needed if we are adding an unique id for the service?

sethvargo: probably not. removed.

.github/workflows/appengine-deploy-it.yml
bharathkkb: Would these run if we modify just the workflow yaml without any changes in `appengine-deploy/**`? For instance if I modify just the `appengine-deploy-it.yml`?

sethvargo: Judging by the commits I've pushed against this branch, yes? Although that's not documented as such.

bharathkkb: @sethvargo I staged an example here, where I modified just the workflow yaml and it did not trigger a build, when I opened a PR.
https://github.com/bharathkkb/test-push-gha/pull/1

bharathkkb: Also looks like only `appengine-deploy Unit` has run for this PR, unless I am missing something. That workflow has `on: [push, pull_request]`.https://github.com/GoogleCloudPlatform/github-actions/blob/ba52be44d0a10d6d8b9e2b6046df628b19f06af6/.github/workflows/appengine-deploy.yml#L2
I think the expected behaviour should be if any of the `action-dir/**`  or corresponding workflow yamls changes, both IT and UT should run.

sethvargo: Okay, updated.

.github/workflows/appengine-deploy-it.yml
bharathkkb: seems like GAE cannot have underscores in the service name and one of the `github.job` is `b64_json`.
https://github.com/GoogleCloudPlatform/github-actions/pull/123/checks?check_run_id=747208506

.github/workflows/appengine-deploy-it.yml
bharathkkb: Does it make sense to add a retry with `--retry-delay`? This deploy looks successful.
https://github.com/GoogleCloudPlatform/github-actions/pull/123/checks?check_run_id=747208477

sethvargo: Done

.github/workflows/appengine-deploy-it.yml
sethvargo: Done

sethvargo:
Well that was certainly an adventure of edge cases. @averikitsch @bharathkkb please re-review when you get a chance.

sethvargo:
Thanks 😄

PR #140, opened 2020-07-16
Added action to get an IAP authentication token.

fcoclavero:
The token can then be used to authenticate to IAP secured resources. I've included tests and documentation.


fcoclavero:
Hey! Just checking in. I'd be glad to help with any doubts you have, or modify anything you see fit 😉.

get-iap-token/action.yaml
```diff
@@ -0,0 +1,24 @@
...
+description: |-
+  Get a JWT that can be used to authenticate API calls behind a Goocle Cloud
+  Identity Aware Proxy (IAP).
+
+inputs:
+  iap_oauth_client_id: # input id
+    description: "The client ID for the IAP OAuth client."
+    required: true
+
+  service_account_key: # input id
```
bharathkkb, line 14: nit:
```suggestion
  credentials: # input id
```

get-iap-token/index.js
```diff
@@ -0,0 +1,52 @@
+const axios = require('axios');
```
bharathkkb, line 1: do we need this file? I see the src as well as dist dirs.

get-iap-token/src/main.ts
```diff
@@ -0,0 +1,24 @@
+import * as core from '@actions/core';
+import { GoogleToken } from 'gtoken';
+import { getCredentials } from './credentials';
+
+// Retrieve input.
```
bharathkkb, line 5: Is there a reason why the the operation is not async?
https://github.com/actions/typescript-action#change-the-code

get-iap-token/src/main.ts
```diff
@@ -0,0 +1,24 @@
...
+const IAPOAuthClientID = core.getInput('iap_oauth_client_id');
+const credentials = getCredentials();
```
bharathkkb, line 7: super nit: we seem to be following a convention where we grab all the inputs first. This way its easier to test helpers.

fcoclavero:
> > @bharathkk thanks a lot for the review! I've modified the changed files according to your comments.
> 
> you have got the wrong alias. can you please untag me so that I don't get updates on this PR?

Sorry!

get-iap-token/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "get-iap-token",
+  "version": "0.1.0",
+  "description": "GitHub action for getting a JWT that can be used to authenticate API calls behind a Goocle Cloud Identity Aware Proxy (IAP).",
```
bharathkkb, line 4:
```suggestion
  "description": "GitHub action for getting a JWT that can be used to authenticate API calls behind a Google Cloud Identity Aware Proxy (IAP).",
```

.github/workflows/get-iap-token-it.yml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          export_default_credentials: true
+          credentials: ${{ secrets.SERVICE_ACCOUNT_KEY_B64 }}
+      - id: iap-token
+        name: get IAP token
+        uses: ./get-iap-token
+        with:
+          iap_oauth_client_id: ${{ secrets.IAP_OAUTH_CLIENT_ID }}
+      - name: get output
+        run: echo '${{ steps.iap-token.outputs.token }}'
```
averikitsch, line 25: We shouldn't be printing sensitive info. We need to test this in a different way

peteos123 (approved):

sethvargo:
@averikitsch @bharathkkb I think we might be able to close this out and point folks to https://github.com/google-github-actions/auth#generating-id-tokens instead. You should be able to mint the JWT using the auth action. Even if that's not possible, I think this probably belongs in that repo instead.

PR #141, opened 2020-07-16
docs(deliverables): multi yaml example

ndeitch:


ndeitch:
@googlebot I signed it!

PR #149, opened 2020-07-22
Revert "Use RUNNER_TEMP to export credentials"

sethvargo:
Reverts GoogleCloudPlatform/github-actions#148

As I kinda suspected but couldn't remember, there's a reason we use the action's temporary directory.

/cc @travisgroth @ViacheslavKudinov

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,3 +1,0 @@
-{
-  "lockfileVersion": 1
-}
--- a/setup-gcloud/README.md
+++ b/setup-gcloud/README.md
@@ -40,6 +40,9 @@ It does the following:
 
     **IMPORTANT!** Exporting default credentials requires `actions/checkout@v2`.
     The `v1` tag is not supported and will not work.
+    
+    **WARNING!** This persists the application credentials inside the workspace. If
+    you are building packages or artifacts, be sure to exclude the credentials!
 
 ## Prerequisites
 
--- a/setup-gcloud/README.md
+++ b/setup-gcloud/README.md
@@ -41,10 +41,6 @@ It does the following:
     **IMPORTANT!** Exporting default credentials requires `actions/checkout@v2`.
     The `v1` tag is not supported and will not work.
 
-    **WARNING!** This persists the credentials to a temporary path on disk. If
-    you are building packages, ensure you do **not** use `RUNNER_TEMP` or you
-    risk including the service account credential in your build.
-
 ## Prerequisites
 
 * This action requires [Python](https://www.python.org/) 2.7.9 or later to be installed on the environment.
--- a/setup-gcloud/dist/index.js
+++ b/setup-gcloud/dist/index.js
@@ -358,9 +358,6 @@ Object.defineProperty(exports, "__esModule", { value: true });
 const os = __webpack_require__(87);
 const events = __webpack_require__(614);
 const child = __webpack_require__(129);
-const path = __webpack_require__(622);
-const io = __webpack_require__(1);
-const ioUtil = __webpack_require__(672);
 /* eslint-disable @typescript-eslint/unbound-method */
 const IS_WINDOWS = process.platform === 'win32';
 /*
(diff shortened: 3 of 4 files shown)


travisgroth:
Is there no other place these can be kept?  Does `HOME` get mounted in docker steps?  https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#filesystems-on-github-hosted-runners

It'd be nice if there was a full reference on the different env vars and how they work out in VM vs Container.

ViacheslavKudinov:
> Is there no other place these can be kept? Does `HOME` get mounted in docker steps? https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#filesystems-on-github-hosted-runners
> 
> It'd be nice if there was a full reference on the different env vars and how they work out in VM vs Container.

It how looks default run of Docker based GH action:
`/usr/bin/docker run --name c20191433d3ea2a95661c0f83a8e9e741a17_42b255 --label 87c201 --workdir /github/workspace --rm -e GOOGLE_CREDENTIALS -e TF_VAR_sealed_secrets_master_key -e TF_VAR_sealed_secrets_master_key_crt -e TF_VAR_crypter_decrypter_service_account_key -e TF_WORKING_DIR -e TF_ACTIONS_VERSION -e TF_WORKSPACE -e CLOUDSDK_METRICS_ENVIRONMENT -e GOOGLE_APPLICATION_CREDENTIALS -e INPUT_TF_ACTIONS_VERSION -e INPUT_TF_ACTIONS_SUBCOMMAND -e INPUT_TF_ACTIONS_WORKING_DIR -e INPUT_TF_ACTIONS_CLI_CREDENTIALS_HOSTNAME -e INPUT_TF_ACTIONS_CLI_CREDENTIALS_TOKEN -e INPUT_TF_ACTIONS_COMMENT -e INPUT_TF_ACTIONS_FMT_WRITE -e INPUT_TF_ACTIONS_GCLOUD_SA_KEY -e HOME -e GITHUB_JOB -e GITHUB_REF -e GITHUB_SHA -e GITHUB_REPOSITORY -e GITHUB_REPOSITORY_OWNER -e GITHUB_RUN_ID -e GITHUB_RUN_NUMBER -e GITHUB_ACTOR -e GITHUB_WORKFLOW -e GITHUB_HEAD_REF -e GITHUB_BASE_REF -e GITHUB_EVENT_NAME -e GITHUB_SERVER_URL -e GITHUB_API_URL -e GITHUB_GRAPHQL_URL -e GITHUB_WORKSPACE -e GITHUB_ACTION -e GITHUB_EVENT_PATH -e RUNNER_OS -e RUNNER_TOOL_CACHE -e RUNNER_TEMP -e RUNNER_WORKSPACE -e ACTIONS_RUNTIME_URL -e ACTIONS_RUNTIME_TOKEN -e ACTIONS_CACHE_URL -e GITHUB_ACTIONS=true -e CI=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/runner/work/_temp/_github_home":"/github/home" -v "/home/runner/work/_temp/_github_workflow":"/github/workflow" -v "/home/runner/work/dpe-iac/dpe-iac":"/github/workspace" 87c201:91433d3ea2a95661c0f83a8e9e741a17` 
in theory we can use **option** section of workflow file to fix our case and define mount of volumes like in this example [https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-15](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-15).

Maybe, the way to use one more input param for this action and to set either RUNNER_TEMP or GITHUB_WORKSPACE as a "workspace" also will be fine.

sethvargo:
> in theory we can use option section of workflow file to fix our case and define mount of volumes like in this example https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-15.

Sure, but that's a breaking change. I think we should revert this. If we want to introduce that change, it should be in a new tagged version so people can upgrade safely.

PR #152, opened 2020-07-28
Allow trimming secret values

podollb:
Optionally trim value of secrets as they are pulled from the secret manager.

### DIFF
--- a/get-secretmanager-secrets/src/client.ts
+++ b/get-secretmanager-secrets/src/client.ts
@@ -25,6 +25,7 @@ import { GoogleAuth } from 'google-auth-library';
 type ClientOptions = {
   credentials?: string;
   endpoint?: string;
+  trim?: boolean;
 };
 
 /**
@@ -41,9 +42,11 @@ export class Client {
 
   readonly auth: GoogleAuth;
   readonly endpoint: string;
+  readonly trim: boolean;
 
   constructor(opts?: ClientOptions) {
     this.endpoint = opts?.endpoint || this.defaultEndpoint;
+    this.trim = opts?.trim || false;
 
     if (opts?.credentials) {
       // If the credentials are not JSON, they are probably base64-encoded. Even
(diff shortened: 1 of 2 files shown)


sethvargo:
Can you explain more why you need to trim values? Why not just exclude the newline characters when you create the secret?

podollb:
I'm not sure if it belongs in this action (it probably doesn't); I just made the draft PR as a way to start a small dialog.

The only reason being...we've been bitten twice with a secret that had an extra trailing whitespace.

podollb:
And bitten once with a trailing newline.

sethvargo:
It's my belief that clients should consume the secret as-is. How are you creating secrets, and is there something you could do (e.g. linter, etc) to prevent this during creation?

podollb:
Generally that is also my belief. In my current use-case, the secrets are being managed (created/updated) by numerous people using the GCP console Secret Manager (i.e. so it's just user error when cutting, pasting, and then saving the secret). 

It sounds like I'll probably need to write a helper tool/script for them to use, when creating/updating secrets (that does the verification/trimming upon saving). Or I could do the trimming in my github action workflow to trim the values once I extract them (with this action).

sethvargo:
Yea, I'm 👎 on munging the data here. I think it's better to fix it upstream when you create the secret.

podollb:
I forgot to ask... Is there a way with the existing action, by way of piping or wrapping the value in some way, to trim (with a tool like `tr`, `sed`, etc.)?

If I grabbed the secret (that had a trailing whitespace in it):

```
      - name: fetch secrets from secret manager so they are available for this deploy
        id: secrets
        uses: GoogleCloudPlatform/github-actions/get-secretmanager-secrets@master
        with:
          credentials: ${{ secrets.SERVICE_ACCOUNT_JSON_KEY }}
          secrets: |-
            SLACK_WEBHOOK:${{ env.GCP_PROJECT_ID }}/SLACK_WEBHOOK
```

And then assuming later on it's used/referenced:

```
      - name: slack notification for failure
        if: failure()
        uses: rtCamp/action-slack-notify@v2.0.2
        env:
          SLACK_WEBHOOK: ${{ steps.secrets.outputs.SLACK_WEBHOOK }}
```

PR #167, opened 2020-09-02
Fix appengine-deploy action to read from image_url param correctly

tsutsu:
The action.yml specifies an `image_url` param, but `main.ts` is looking for an `image-url` parameter instead, and so any `image_url` passed in gets silently dropped. (And `image-url` isn't grammatically valid as an action-parameter identifier, so you can't just pass that in, either.)

I fixed `main.ts`, and ran `npm run build` to regenerate `dist/`. Also updated README.md to refer to the parameter as `image_url`.


averikitsch (approved):

averikitsch:
Thank you

PR #170, opened 2020-09-04
[DRAFT] feat: Test container builds

bharathkkb:
This enables running NCC within a container to have reproducible builds.
Changes:

- `npm run build` now builds a container and runs the build for the respective action within that container
- added an additional compile check that ensures the SHAs of artifacts match


bharathkkb:
/cc @bcoe this is what I had in mind 
PTAL

bharathkkb:
superseded via #190

PR #173, opened 2020-09-12
Use last matched URL

garsue:
The first URL is always matched with "https://cloud.google.com".


garsue:
@googlebot I signed it!

averikitsch:
Integration tests did not run.

averikitsch:
Hi @garsue thanks for the fix. Can you add a check to the [integration tests](https://github.com/GoogleCloudPlatform/github-actions/blob/master/.github/workflows/appengine-deploy-it.yml) to make sure the url contains `appspot.com` or that is doesn't contain cloud.google.com? Thanks!

averikitsch:
Merging into a test branch

averikitsch:
Based on https://github.com/GoogleCloudPlatform/github-actions/actions/runs/325078654 this doesn't work.

PR #178, opened 2020-09-24
[draft] test jit dist compilation for integration tests

bharathkkb:
- removed `dist/index.js` for SM action as an example
- added steps to build before running integration tests


bharathkkb:
implemented via #190

PR #190, opened 2020-10-06
feat: remove dist, add release-please

bharathkkb:


averikitsch:
If we remove dist/ before we run release-please this will break the actions

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,43 @@
...
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+        max-parallel: 1
+        matrix: 
+          actions: ["setupGcloudSDK", "appengine-deploy","deploy-cloudrun", "setup-gcloud", "get-secretmanager-secrets", "upload-cloud-storage"]
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+           git pull
```
averikitsch, line 16: nit: do this in one line

.github/workflows/appengine-deploy-it.yml
```diff
@@ -31,6 +31,18 @@ jobs:
         });
         server.listen(process.env.PORT || 8080);
         EOF
+    - id: build dependency
+      name: Build dependency
+      working-directory: setupGcloudSDK
```
averikitsch, line 36: Should we add this to the run on push to path at the top of this file?

.github/workflows/appengine-deploy-it.yml
```diff
@@ -31,6 +31,18 @@ jobs:
...
+      run: |-
+        npm install
+        npm run build
+    - id: build dependency
+      name: Build dependency
+      working-directory: setupGcloudSDK
```
averikitsch, line 42: appengine-deploy?

.github/workflows/deploy-cloudrun-it.yml
```diff
@@ -164,7 +200,7 @@ jobs:
         node-version: 12.x
     - run: npm install
       working-directory: ./deploy-cloudrun
-    - uses: ./setup-gcloud # Set up ADC to make authenticated request to service # Set up ADC to make authenticated request to service
+    - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master # Set up ADC to make authenticated request to service # Set up ADC to make authenticated request to service
```
averikitsch, line 203: comment duplicated

.github/workflows/setup-gcloud-it.yml
```diff
@@ -16,7 +16,12 @@ jobs:
         operating-system: [ubuntu-latest, windows-latest, macos-latest]
     steps:
     - uses: actions/checkout@v2
-
+    - id: build
```
averikitsch, line 19: Don't we need to build the setupGcloudSdk here too?

bharathkkb:
> If we remove dist/ before we run release-please this will break the actions

@averikitsch so my understanding is that when we merge this to main because of https://github.com/GoogleCloudPlatform/github-actions/blob/e829c1b7b9c2f632d89189668ec8348f5270bfc3/.github/workflows/release-please.yml#L5 this should be triggered and will proceed to recreate those dists and commit to main, so yes technically we will be broken on main for a couple of minutes (or longer if we run into any errors during dist build on main). I will do a final test after I address the comments above with a fresh fork.

bharathkkb:
Oh I think I get what you meant, dont remove existing dist, let it autogenerate and commit instead. That would indeed work better, the reason I removed it here is because I was testing that the JIT build for int tests do work. I will re add them back.

averikitsch:
> Oh I think I get what you meant, dont remove existing dist, let it autogenerate and commit instead. That would indeed work better, the reason I removed it here is because I was testing that the JIT build for int tests do work. I will re add them back.

Yes this sounds good. Note to self, leave better comments.

.github/workflows/appengine-deploy-it.yml
bharathkkb: Did you mean adding `'setupGcloudSDK**'` so that we test `appengine-deploy` for any change in`setupGcloudSDK`?

.github/workflows/appengine-deploy-it.yml
bharathkkb: yup good catch

.github/workflows/setup-gcloud-it.yml
bharathkkb: ah yes. This was correctly surfaced when I pushed the first commit https://github.com/GoogleCloudPlatform/github-actions/runs/1217429019?check_suite_focus=true but what not triggered for subsequent commits because of push-path. Multi repo should catch these 😅

.github/workflows/appengine-deploy-it.yml
bharathkkb: done

.github/workflows/deploy-cloudrun-it.yml
bharathkkb: fixed

.github/workflows/setup-gcloud-it.yml
bharathkkb: fixed

.github/workflows/release-please.yml
bharathkkb: done

.github/workflows/appengine-deploy-it.yml
bharathkkb: done

bharathkkb:
@averikitsch did another build release test here with current workflow https://github.com/bharathkkb/github-actions/actions/runs/298346191

averikitsch (approved):

PR #214, opened 2020-10-24
Only run test when ADC is found

averikitsch:


deploy-cloudrun/tests/unit/cloudRun.test.ts
```diff
@@ -38,6 +38,9 @@ describe('CloudRun', function() {
   });
 
   it('initializes with ADC', async function() {
+    if (process.env.GOOGLE_APPLICATION_CREDENTIALS) {
```
bharathkkb, line 41: This makes sense, but in this particular case as this is not a fork, it should be able to run with ADC as we do `export_default_credentials` prior to npm test?
It seems to skip in the latest run
https://github.com/GoogleCloudPlatform/github-actions/runs/1302499186#step:7:18

bharathkkb: Ah I see the release-please bot is creating the release PR from fork and that is failing the test.

bharathkkb (approved):
LGTM

PR #220, opened 2020-10-30
chore: add notices to move to new repo

bharathkkb:
- Adds notices in source code and readme to inform users of upcoming repo changes

TODO:
- [x] individual action repos


appengine-deploy/src/main.ts
```diff
@@ -21,7 +21,9 @@ import * as setupGcloud from '../../setupGcloudSDK/dist/index';
 async function run(): Promise<void> {
   try {
     // Add notice
-    core.warning('Thank you for using appengine-deploy Action. GoogleCloudPlatform/github-actions/appengine-deploy has been deprecated, please switch to google-github-actions/appengine-deploy.');
+    core.warning(
+      'Thank you for using appengine-deploy Action. GoogleCloudPlatform/github-actions/appengine-deploy has been deprecated, please switch to google-github-actions/appengine-deploy.',
```
averikitsch, line 25: For consistency, I want to update this to `deploy-appengine`

averikitsch (approved):

appengine-deploy/src/main.ts
bharathkkb: Done, should `cloudfunctions` in `deploy-cloud-functions` be one word or should I keep it as is?

bharathkkb: as discussed offline keeping as `deploy-cloud-functions` for readability

averikitsch (approved):

bharathkkb:
New changes include
- regenerated dists for all repos with notices
- I realized that the initial approach wont work for `setup-gcloud` as we have not top leveled `setup-gcloud` action yet.


We can either top level `setup-gcloud` action to `/` now and freeze code inside `/setup-gcloud` or what I have currently done is: 
- generated a `dist-frozen` which has deprecation notice and then removed it from src here https://github.com/google-github-actions/setup-gcloud/pull/220/commits/2dfd1c5a216e923d96490cafdf68aa9a42375beb
- targeted `setup-gcloud/action.yml` to point at `dist-frozen/index.js` so that anyone using `GoogleCloudPlatform/github-actions/setup-gcloud` will get the core.info notice.
- targeted all setup-gcloud IT to use `/action.yml` which points to `/setup-gcloud/dist/index.js` in https://github.com/google-github-actions/setup-gcloud/pull/220/commits/96bcc68ff9404ef15940e944e1640c40343e424c which does not have the notice
- users who reference the new`google-github-actions/setup-gcloud` will use the `/action.yml` which inturn will use the `/setup-gcloud/dist/index.js`
- long term when we are ready to delete old actions, all we will need to do is target `/action.yml` to point to `/dist/index.js` instead of `/setup-gcloud/dist/index.js`

I think top leveling might be easier but it may look messy as we have to maintain the 6 other action directories for a while.

PR #227, opened 2020-11-13
chore: remove builds and releases for other actions

bharathkkb:
- remove builds for other actions except setup-gcloudm setupGcloudSDK
- remove releases for other actions except setup-gcloud


averikitsch (approved):
This looks good. We can also structure this similarly to `deploy-appengine` were setupGcloudSDK is not a separate build

PR #237, opened 2020-11-16
fix: action.yml description length

bharathkkb:


action.yml
```diff
@@ -15,8 +15,8 @@
 name: Set up gcloud Cloud SDK environment
 author: GoogleCloudPlatform
 description: |-
-  Downloads, installs, and configures a gcloud Cloud SDK environment for the
-  worker. This will add the `gcloud` CLI command to the worker's $PATH.
+  Downloads, installs, and configures a gcloud Cloud SDK environment.
```
averikitsch, line 18:
```suggestion
  Downloads, installs, and configures the Cloud SDK. 
```

averikitsch (approved):

action.yml
bharathkkb: Changed as discussed offline

PR #241, opened 2020-11-17
docs: update README

averikitsch:


README.md
```diff
@@ -12,43 +12,73 @@
...
+* [Inputs](#inputs)
+* [Example Workflows](#example-workflows)
+
+## Usage
+
+```yaml
+- name: Set up Cloud SDK
+  uses: google-github-actions/setup-gcloud@master
+  with:
+    version: '290.0.1'
```
bharathkkb, line 36: nit: I think we should remove the default version pinning, so that if someone is copying this code block they get latest by default.

README.md
```diff
@@ -12,43 +12,73 @@
...
+
+- name: Use gcloud CLI
+  run: gcloud info
+```
 
-* [Google Kubernetes Engine](./example-workflows/gke/README.md): An example workflow that uses [GitHub Actions][github-action] to deploy a static website to an existing [Google Kubernetes Engine](https://cloud.google.com/kubernetes-engine/) cluster.
+## Inputs
 
-* [Cloud Run](./example-workflows/cloud-run/README.md): An example workflow that uses [GitHub Actions][github-action] to build and deploy a container to [Cloud Run](https://cloud.google.com/run/).
+| Name          | Requirement | Default | Description |
```
bharathkkb, line 47: This is awesome, we should do it for all the other actions. Is there a tool for doing this?

README.md
```diff
@@ -12,43 +12,73 @@
...
 
-# Google Cloud Platform: github-actions
+# `setup-gcloud` GitHub Action
 
-This repository contains a library of [Github Actions](https://github.com/actions) providing functionality for working with [Google Cloud Platform](http://cloud.google.com/).
+Configures the [Google Cloud SDK][sdk] in the GitHub Actions environment. The Google Cloud SDK includes both the [gcloud][gcloud] and
+[gsutil][gsutil] binaries.
 
-## Available Actions
+Find other Google Cloud GitHub Actions in [`google-github-actions`](https://github.com/google-github-actions).
```
bharathkkb, line 20: Since majority of our users may use setup-gcloud, can we surface some of the other actions here so they can make a decision of setup-gcloud vs native action.

bharathkkb (approved):
LGTM
Some minor comments

README.md
averikitsch: Good! I thought this was more readable. I just copied and pasted into a markdown table, but there are tools like https://tableconvert.com/

README.md
averikitsch: Do we like the actions listed in this way?

bharathkkb: Yup those look great and increased visibility!

README.md
bharathkkb: gotcha, I am thinking of writing a small script that can do this for use. I will update in a PR.

bharathkkb:
@averikitsch feel free to merge when ready

PR #250, opened 2020-12-18
Update action name

lewislbr:
Updated the action name to use the new format in all instances expect those related to the deprecation notice.


lewislbr:
@googlebot I signed it!

bharathkkb (approved):
Thanks for the PR! LGTM; Some tests seems to be failing due to secret access.

PR #254, opened 2020-12-21
fix: point users to specific example workflow

averikitsch:


example-workflows/cloud-run/README.md
```diff
@@ -1,21 +1,25 @@
...
+_**Checkout the [`deploy-cloudrun` action](https://github.com/google-github-actions/deploy-cloudrun) and [example workflows](https://github.com/google-github-actions/deploy-cloudrun/README.md#example-workflows)
+for a specialized implementation.**_
 
 This code is intended to be an _example_. You will likely need to change or
 update values to match your setup.
 
 ## Workflow description
 
-For pushes to the `master` branch, this workflow will:
+For pushes to the `example` branch, this workflow will:
```
bharathkkb, line 13: For triggering don't we need to copy the workflow from `example-workflows/cloud-run/cloud-run.yml ` to `.github/workflows`?

example-workflows/cloud-run/cloud-run.yml
```diff
@@ -0,0 +1,53 @@
...
+      with:
+        project_id: ${{ env.PROJECT_ID }}
+        service_account_key: ${{ secrets.GCP_SA_KEY }}
+
+    - name: Authorize Docker push
+      run: gcloud auth configure-docker
+
+    - name: Build and Push Container
+      run: |-
+        docker build -t gcr.io/${{ env.PROJECT_ID }}/${{ env.SERVICE }}:${{  github.sha }} .
```
bharathkkb, line 44: Looks like `example-workflows/cloud-run/Dockerfile` was removed?

bharathkkb (commented):
Looks like some of the files were removed which may prevent the workflow from being functional out of the box?

averikitsch:
> Looks like some of the files were removed which may prevent the workflow from being functional out of the box?

I'm debating here not having the full example app but just showing the workflow file. I am trying to push people towards the specific actions. The "workflow description" just describes what is happening in the file.

bharathkkb (approved):
@averikitsch I see, in that case LGTM

PR #313, opened 2021-05-25
Update gke.yml

ChristophDietrich:
Adding missing $ to var of PROJECT_ID in deploy step


ChristophDietrich:
@googlebot I signed it!

PR #331, opened 2021-06-26
lingering org transition cleanup

tequilarista:
Deleted old non-setup-gcloud actions code that's since been migrated to their own repos within the new org.  Cleaned out superfluous workflows related to same.

Uplifted setup-cloud/setup-cloud/* to setup-gcloud/.

Uplifted lint config files to top level and cleaned up changes in setupGcloudSDK related to this.


action.yml
```diff
@@ -13,10 +13,10 @@
 # limitations under the License.
 
 name: Set up gcloud Cloud SDK environment
-author: Google LLC
+author: GoogleCloudPlatform
```
averikitsch, line 16: Changes to this file need to be reverted (this is for the deprecate code)

setup-gcloud/README.md
```diff
@@ -25,74 +25,3 @@ steps:
 +  uses: google-github-actions/setup-gcloud@master
```
averikitsch, line 25: I think we can go ahead and delete this file. @bharathkkb do you think we should add a note in the readme about the deprecated actions?

bharathkkb: Yeah I think we should add a readme note stating these are not deprecated, but rather live elsewhere.

action.yml
```diff
@@ -70,4 +69,4 @@ inputs:
 
 runs:
   using: node12
-  main: setup-gcloud/dist/index.js
+  main: dist-frozen/index.js
```
bharathkkb, line 72: we can remove `dist-frozen`. this should point to `dist/index.js`

action.yml
tequilarista: reverted.

action.yml
tequilarista: done

.github/workflows/setup-gcloud-it.yml
```diff
@@ -3,7 +3,7 @@ name: setup-gcloud Integration
 on:
   push:
     paths:
-    - 'setup-gcloud/**'
+    - './**'
     - '.github/workflows/setup-gcloud*'
```
bharathkkb, line 7: nit: Now that we have top leveled this, we can use
```yaml
  push:
    branches:
    - 'main'
  pull_request:
```
for parity with other actions test workflows
https://github.com/google-github-actions/deploy-cloudrun/blob/c8ab46dee1f5dad5d4ec6e92519bf90fa9318923/.github/workflows/deploy-cloudrun-it.yml#L3

setup-gcloud/README.md
bharathkkb: @tequilarista I am thinking we can move the deprecation notice to main root readme for increased visibility and get rid of `setup-gcloud` dir completely?

bharathkkb (changes requested):
Some minor changes but looks good overall. Thanks for taking care of this!

- There seems to be a diff for`setup-gcloud/dist/index.js → dist/index.js`. It maybe because it was built locally and added in a commit. We should remove this checked in file and just rename it without changing source code, our build workflow should generate a new version of this.
- We should update [build/release workflow](https://github.com/google-github-actions/setup-gcloud/blob/be078c0cd4467957aec54a8ac035fbe9bbdc0a0c/.github/workflows/release-please.yml#L19) as`setup-gcloud` dir no longer exists. I propose we just  build/release setup-gcloud as @averikitsch is splitting out the SDK anyway.

tsconfig.json
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Google LLC
+ * Copyright 2019 Google LLC
```
iennae, line 2: why this change? seems weird.

tequilarista: came over with one of the files moves, I can put it back

setup-gcloud/README.md
tequilarista: makes sense, will update

bharathkkb (approved):
LGTM - did a quick test [here](https://github.com/bharathkkb/test-setup-gcloud/blob/main/.github/workflows/gcloud.yml) for setup-gcloud

The dist changes are there but I think we can just let it get recreated after merging this PR.

/cc @averikitsch for another 👀

averikitsch (approved):

PR #334, opened 2021-07-01
chore: update dependencies from vulnerable versions

averikitsch:


bharathkkb:
@averikitsch we currently don't have automation to auto build these https://github.com/google-github-actions/setup-gcloud/blob/0aa92bcd70fc0929f052b47276d5e61f793c384b/.github/workflows/release-please.yml#L14 Not sure if we should manually build these due to https://github.com/google-github-actions/setup-gcloud/pull/331

averikitsch:
Right.. trying to slap a bandaid on the security alerts but we should probably just push #331 through.

PR #385, opened 2021-11-09
fixing arm64 gcloud sdk conventions naming

komondor:


sethvargo:
Hi @komondor 

This diff is fairly substantial for a small change.

PR #387, opened 2021-11-10
Add support for arm

sethvargo:
Node names this "arm64", but the Cloud SDK binary is just "arm".

Moved this to a map to avoid lots of if statements as it expands in the future.


sethvargo:
Closes https://github.com/google-github-actions/setup-gcloud/pull/386

bharathkkb (changes requested):
@sethvargo We should add this to https://github.com/google-github-actions/setup-cloud-sdk. We are planning on using that in setup-gcloud and removing the SDK here.

PR #397, opened 2021-11-23
feat: update to use setup-cloud-sdk package

averikitsch:


package.json
```diff
@@ -28,6 +28,7 @@
     "@actions/exec": "^1.0.3",
     "@actions/io": "^1.1.1",
     "@actions/tool-cache": "^1.7.1",
+    "@google-github-actions/setup-cloud-sdk": "^0.1.1",
```
bharathkkb, line 31: Looks like we can also clean up some of the other unused deps

src/setup-gcloud.ts
```diff
@@ -90,6 +95,10 @@ export async function run(): Promise<void> {
       core.info('Successfully exported Default Application Credentials');
     }
   } catch (error) {
-    core.setFailed(error.message);
+    if (error instanceof Error) {
+      core.setFailed(error.message);
+    } else {
+      core.setFailed("'setup-gcloud' failed to be installed.");
+    }
```
bharathkkb, line 102: Cant we just `core.setFailed('setup-gcloud' failed to be installed: ${errror});`

bharathkkb (approved):

PR #403, opened 2021-11-30
Upgrade to latest lockfile version

sethvargo:
I'm working on something else, but this is resulting in a large diff, so I'm doing it separately.


sethvargo:
This will also theoretically make dependabot less noisy: https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json

bharathkkb (approved):

PR #405, opened 2021-11-30
Export credentials into RUNNER_TEMP instead of GITHUB_WORKSPACE

sethvargo:
This will ensure the credentials do no persist beyond the run, but are still available to all other steps in the job.

Also a few small fixes:

- Remove deprecated prettier config option
- Drop dependency on uuid and use crypto instead for generating filenames
- Write the credentials file securely

Fixes https://github.com/google-github-actions/setup-gcloud/issues/402

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,6 +20,7 @@ module.exports = {
   plugins: ['@typescript-eslint'],
   extends: [
     'eslint:recommended',
+    'plugin:@typescript-eslint/eslint-recommended',
     'plugin:@typescript-eslint/recommended',
     'plugin:prettier/recommended',
   ],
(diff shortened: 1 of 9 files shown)


bharathkkb (approved):

jamiet-msm:
thx for this @sethvargo

jamiet-msm:
@sethvargo We're getting this warning from our use of `google-github-actions/setup-gcloud@master`

> terraform-fmt-validate-runtime
"service_account_key" has been deprecated. Please switch to using google-github-actions/auth which supports both Workload Identity Federation and Service Account Key JSON authentication. For more details, see https://github.com/google-github-actions/setup-gcloud#authorization

We're happy to make the change to google-github-actions/auth for authentication but we don't want to fall foul of the issue that you fixed in this PR. Will this change be reflected if we make the move to google-github-actions/auth ?

cc @marengaz

sethvargo:
@jamiet-msm yea, we're trying to centralize all authentication into "auth". auth already [has this fix](https://github.com/google-github-actions/auth/pull/67).

PR #407, opened 2021-11-30
Run `run ci` instead of `npm install` in release please

sethvargo:
This should ensure the lockfile doesn't change (I think).

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -9,12 +9,16 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - run: git pull
-      - name: install
-        run: npm install
-      - name: build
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yml
```diff
@@ -9,12 +9,16 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - run: git pull
```
sethvargo, line 13: I'm assuming the `git pull` here was to get the entire history. I updated the checkout action to use `fetch-depth:0` which will fetch the entire history.

bharathkkb: Yes IIRC the newer version of this script does use `fetch-depth:0` for other actions

bharathkkb (approved):

Issue #410, opened 2021-12-01
Error while running plan with this GA | Terraform 1.0.8


Sunil-Jacob:
### TL;DR
Getting the below error:

```text
Error: storage.NewClient() failed: dialing: google: error getting credentials using GOOGLE_APPLICATION_CREDENTIALS environment variable: open /home/runner/work/_temp/651d8f6a44067dceb917e2cf: no such file or directory
```

**Expected behavior**
The plan should run successfully

**Observed behavior**
Getting the below error:

```text
Error: storage.NewClient() failed: dialing: google: error getting credentials using GOOGLE_APPLICATION_CREDENTIALS environment variable: open /home/runner/work/_temp/651d8f6a44067dceb917e2cf: no such file or directory
```

### Reproduction

**Action YAML**

```yaml
steps:
      # Checkout the repository to the GitHub Actions runner
      - name: Checkout
        uses: actions/checkout@v2

      - name: Set up Cloud SDK
        uses: google-github-actions/setup-gcloud@master
        with:
          project_id: ${{ secrets.GCP_PROJECT_ID }}
          service_account_key: ${{ secrets.GOOGLE_PRIVATE_KEY }}
          export_default_credentials: true

      - name: Bucket creation DEV
        shell: bash
        run: |
          gsutil ls -b gs://${{ secrets.GCP_PROJECT_NAME}} || (gcloud services enable storage-api.googleapis.com && gsutil mb gs://${{ secrets.GCP_PROJECT_NAME}} && gsutil versioning set on gs://${{ secrets.GCP_PROJECT_NAME}} )
      - name: plan-dev
        uses: dflook/terraform-plan@v1
        env:
          GITHUB_TOKEN : ${{ secrets.GITHUB_TOKEN }}
          TERRAFORM_HTTP_CREDENTIALS: |
            github.com/ebomart=${{ secrets.CI_USER}}:${{ secrets.CI_TOKEN}}
        with:
          path: ./
          var_file: environment/dev/variables.tfvars
          label: dev
          backend_config_file: environment/dev/gcs-bucket.tfvars

```

**Additional information**
Are you running custom workers? Doing something atypical? No

Sunil-Jacob:
Hi Team,

This is working when I used SHA `google-github-actions/setup-gcloud@b8f95eb7d716bf9a715eafd169162b93f69ed520`

aurecchia:
[...]

PR #412, opened 2021-12-01
Revert back to using GITHUB_WORKSPACE for storing credentials

sethvargo:
This reverts the change from using `RUNNER_TEMP` to store exported credentials back to the original value of `GITHUB_WORKSPACE` which many additional comments and tests to ensure there's not a regression in the future 😄 

Fixes GH-410


.github/workflows/setup-gcloud-it.yml
```diff
@@ -4,7 +4,13 @@ on:
   push:
     branches:
     - 'master'
-    pull_request:
```
sethvargo, line 7: I'm not sure what our desired intended behavior was here. I don't think nesting `pull_request` under `push` is valid. I've updated it here to run on all PRs, but I think that will fail since outside contributors won't have access to the secrets. LMK what our desired behavior is here.

.github/workflows/setup-gcloud.yml
```diff
@@ -11,18 +22,23 @@ jobs:
...
+    - uses: 'actions/checkout@v2'
 
-    - name: Set Node.js 12.x
-      uses: actions/setup-node@master
+    - uses: 'actions/setup-node@v2'
       with:
-        node-version: 12.x
+        node-version: '12.x'
+
+    - name: 'npm ci'
```
sethvargo, line 31: Switch from `install` to `ci` so not to generate a lockfile diff

action.yml
```diff
@@ -68,6 +68,14 @@ inputs:
...
+      filesystem upon completion.
+    default: true
+    required: false
+
 runs:
-  using: node12
-  main: dist/index.js
+  using: 'node12'
+  main: 'dist/main/index.js'
+  post: 'dist/post/index.js'
```
sethvargo, line 81: There appears to be no good way to "test" this. I've verified it's working by looking at the log output in our tests.

.github/workflows/setup-gcloud-it.yml
bharathkkb: Yeah this looks like a typo. For external contributions via forks the `github.repository == github.event.pull_request.head.repo.full_name` would be false and should skip those jobs.

src/setup-gcloud.ts
```diff
@@ -87,9 +69,19 @@ export async function run(): Promise<void> {
...
+        // instead of RUNNER_TEMP, because RUNNER_TEMP is not shared with
+        // Docker-based actions on the filesystem. Exporting to GITHUB_WORKSPACE
+        // ensures that the exported credentials are automatically available to
+        // Docker-based actions without user modification.
+        //
+        // This has the unintended side-effect of leaking credentials over time,
+        // because GITHUB_WORKSPACE is not automatically cleaned up on
+        // self-hosted runners. To mitigate this issue, this action defines a
+        // post step to remove any created credentials.
+        const runnerTempDir = process.env.GITHUB_WORKSPACE;
```
bharathkkb, line 82: Should we name it `workspaceDir` since it's not the RUNNER_TEMP dir?

sethvargo: Done

bharathkkb (approved):

yonghaoy:
Hello @sethvargo our issue is not totally resolved by this PR.
Here is our GHA: https://github.com/DataBiosphere/terra-docker/blob/master/.github/workflows/test-terra-jupyter-aou.yml#L55L81
Now we are getting permission denied PermissionError: [Errno 13] Permission denied: '/tmp/credentials.json'
pinning us to an older commit is still working. 
Thanks

sethvargo:
That's likely because the `USER` for your docker container is not mapped back to the system and therefore doesn't have permissions to read the file. You can try setting the user manually `--user "$(id -u):$(id -g)"`.

The fact that this worked previously was a rather large security vulnerability, as any user would have been able to access and read those credentials regardless of their OS permissions.

If you'd like to discuss further, please [open an issue](https://github.com/google-github-actions/setup-gcloud/issues/new/choose) since pull requests aren't great for this type of discussion.

PR #413, opened 2021-12-02
feat: Support WIF creds

bharathkkb:
fixes https://github.com/google-github-actions/setup-gcloud/issues/396

### DIFF
--- a/.github/workflows/setup-gcloud-it.yml
+++ b/.github/workflows/setup-gcloud-it.yml
@@ -177,3 +177,81 @@ jobs:
       with:
         entrypoint: '/bin/sh'
         args: '-euc "test -n "${GOOGLE_APPLICATION_CREDENTIALS}" && test -r "${GOOGLE_APPLICATION_CREDENTIALS}"'
+
+  wif:
+    if: ${{ github.event_name == 'push' || github.repository == github.event.pull_request.head.repo.full_name && github.actor != 'dependabot[bot]' }}
+    name: 'with wif'
+    runs-on: '${{ matrix.os }}'
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+        - 'ubuntu-latest'
+        - 'windows-latest'
+        - 'macos-latest'
+
+    permissions:
+      id-token: 'write'
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: 'actions/setup-node@v2'
+        with:
+          node-version: '12.x'
+
+      - id: build
+        name: Build dist
+        run: |-
+          npm ci
+          npm run build
+
+      - uses: google-github-actions/auth@main
+        with:
+          workload_identity_provider: ${{ secrets.WIF_PROVIDER_NAME }}
+          service_account: ${{ secrets.OIDC_AUTH_SA_EMAIL }}
+
+      - name: 'setup-gcloud'
+        uses: './'
+      
+      - id: 'gcloud'
+        shell: 'bash'
+        run: |-
+          gcloud secrets versions access "latest" --secret "${{ secrets.OIDC_AUTH_TEST_SECRET_NAME }}"
+
+  auth-key:
+    if: ${{ github.event_name == 'push' || github.repository == github.event.pull_request.head.repo.full_name && github.actor != 'dependabot[bot]' }}
+    name: 'with key and auth action'
+    runs-on: '${{ matrix.os }}'
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+        - 'ubuntu-latest'
+        - 'windows-latest'
+        - 'macos-latest'
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: 'actions/setup-node@v2'
+        with:
+          node-version: '12.x'
+
+      - id: build
+        name: Build dist
+        run: |-
+          npm ci
+          npm run build
+
+      - uses: google-github-actions/auth@main
+        with:
+          credentials_json: ${{ secrets.SETUP_GCLOUD_IT_KEY }}
+
+      - name: 'setup-gcloud'
+        uses: './'
+      
+      - id: 'gcloud'
+        shell: 'bash'
+        run: |-
+          gcloud secrets versions access "latest" --secret "${{ secrets.OIDC_AUTH_TEST_SECRET_NAME }}"
--- a/.github/workflows/setup-gcloud-it.yml
+++ b/.github/workflows/setup-gcloud-it.yml
@@ -219,7 +219,7 @@ jobs:
         run: |-
           gcloud secrets versions access "latest" --secret "${{ secrets.OIDC_AUTH_TEST_SECRET_NAME }}"
 
-  auth-key:
+  credentials_json:
     if: ${{ github.event_name == 'push' || github.repository == github.event.pull_request.head.repo.full_name && github.actor != 'dependabot[bot]' }}
     name: 'with key and auth action'
     runs-on: '${{ matrix.os }}'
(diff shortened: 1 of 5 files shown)


src/setup-gcloud.ts
```diff
@@ -55,12 +55,20 @@ export async function run(): Promise<void> {
...
 
     const serviceAccountKey = core.getInput('service_account_key');
-    // If a service account key isn't provided, log an un-authenticated notice
-    if (!serviceAccountKey) {
-      core.info('No credentials provided, skipping authentication');
-      return;
-    } else {
+    // If a service account key is provided, add warning to use google-github-actions/auth
+    if (serviceAccountKey) {
+      core.warning(
```
bharathkkb, line 60: TBD if we want to add this in setup-gcloud. I kept it for consistency with the refactors we did in other actions.

.github/workflows/setup-gcloud-it.yml
```diff
@@ -177,3 +177,81 @@ jobs:
...
+      - name: 'setup-gcloud'
+        uses: './'
+      
+      - id: 'gcloud'
+        shell: 'bash'
+        run: |-
+          gcloud secrets versions access "latest" --secret "${{ secrets.OIDC_AUTH_TEST_SECRET_NAME }}"
+
+  auth-key:
+    if: ${{ github.event_name == 'push' || github.repository == github.event.pull_request.head.repo.full_name && github.actor != 'dependabot[bot]' }}
```
sethvargo, line 223: To consider in a follow-up. I wonder if we can store this conditional as an [env](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#env) and then just reference the `env` in the conditional instead. I think:

```yaml
if: ${{ env.CAN_ACCESS_SECRETS }}
```

might be a little clearer. Alternatively, maybe we could do something with [defaults](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#defaults) to have a default `if`?

Low priority, just thinking out loud.

README.md
```diff
@@ -74,7 +76,7 @@ steps:
...
-| `service_account_key`   | _optional_  | | The service account key which will be used for authentication credentials. This key should be [created](https://cloud.google.com/iam/docs/creating-managing-service-account-keys) and stored as a [secret](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets). It can be encoded as a [Base64](https://en.wikipedia.org/wiki/Base64) string or as JSON. |
+| `service_account_key`   | _optional_  | | (**Deprecated**) This input is deprecated. See [auth section](https://github.com/google-github-actions/setup-gcloud#via-google-github-actionsauth) for more details. The service account key which will be used for authentication credentials. This key should be [created](https://cloud.google.com/iam/docs/creating-managing-service-account-keys) and stored as a [secret](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets). It can be encoded as a [Base64](https://en.wikipedia.org/wiki/Base64) string or as JSON. |
```
sethvargo, line 79: Since this is deprecated, can we move it to to the bottom of the table?

README.md
```diff
@@ -95,9 +97,61 @@ code to [App Engine](https://cloud.google.com/appengine), a fully managed server
 * [Cloud Build](./example-workflows/cloud-build/README.md): An example workflow that uses GitHub Actions to build a container image with [Cloud Build](https://cloud.google.com/cloud-build).
 
 
-## Sharing Credentials
+## Authorization
+
+There are a few ways to authenticate this action.
```
sethvargo, line 102:
```suggestion
This action installs the Cloud SDK (`gcloud`). To configure its authentication to Google Cloud, use the `google-github-actions/auth` action. You can authenticate via:

- **Workload Identity Federation** (preferred):

    INSERT_YAML_EXAMPLE

- **Service Account Key JSON**:

    INSERT_YAML_EXAMPLE
    
- **Application Default Credentials**, if and only if you are using self-hosted runners that are hosted on Google Cloud Platform. The Cloud SDK will automatically authenticate using the machine credentials in this case:

    INSERT_YAML_HERE
```

src/setup-gcloud.ts
```diff
@@ -55,12 +55,20 @@ export async function run(): Promise<void> {
...
+        '"serviceAccountKey" input has been deprecated. ' +
```
sethvargo, line 61:
```suggestion
        '"service_account_key" has been deprecated. ' +
```

src/setup-gcloud.ts
```diff
@@ -55,12 +55,20 @@ export async function run(): Promise<void> {
...
+          'Please switch to using google-github-actions/auth which supports both Workload Identity Federation and JSON Key authentication. ' +
```
sethvargo, line 62:
```suggestion
          'Please switch to using google-github-actions/auth which supports both Workload Identity Federation and Service Account Key JSON authentication. ' +
```

src/setup-gcloud.ts
```diff
@@ -55,12 +55,20 @@ export async function run(): Promise<void> {
...
+          'For more details, see https://github.com/google-github-actions/setup-gcloud#authorization',
+      );
+    }
+
+    // Either serviceAccountKey or GOOGLE_GHA_CREDS_PATH env var required
+    if (serviceAccountKey || process.env.GOOGLE_GHA_CREDS_PATH) {
       await authenticateGcloudSDK(serviceAccountKey);
+    } else {
+      core.info('No credentials detected, skipping authentication');
```
sethvargo, line 71: Maybe make this `debug`? I'm worried people will see this log message and think they did something incorrect

.github/workflows/setup-gcloud-it.yml
```diff
@@ -177,3 +177,81 @@ jobs:
...
+
+      - name: 'setup-gcloud'
+        uses: './'
+      
+      - id: 'gcloud'
+        shell: 'bash'
+        run: |-
+          gcloud secrets versions access "latest" --secret "${{ secrets.OIDC_AUTH_TEST_SECRET_NAME }}"
+
+  auth-key:
```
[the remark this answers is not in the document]
bharathkkb: done

.github/workflows/setup-gcloud-it.yml
bharathkkb: Yeah its getting cumbersome. Opened https://github.com/google-github-actions/setup-gcloud/issues/414 to track for followup.

README.md
bharathkkb: moved

src/setup-gcloud.ts
bharathkkb: done

src/setup-gcloud.ts
bharathkkb: updated

src/setup-gcloud.ts
bharathkkb: Historically this [has been](https://github.com/google-github-actions/setup-gcloud/blob/94faef24d2af9aed96d8838c7252ae4f7165dc76/src/setup-gcloud.ts#L60) info lvl so folks should be used to it now. I also think it's easier for users to identify/self service issues like empty GH secrets if we log at info level as opposed to debug which requires setting ACTIONS_STEP_DEBUG.

sethvargo (approved):

steveoh:
Will you be sticking with the 0 based versioning with the next release?

bharathkkb:
@steveoh yes it will be 0.3.0 https://github.com/google-github-actions/setup-gcloud/pull/338 for tracking. We will likely give the new features some soak time before cutting a 1.0.

PR #426, opened 2021-12-07
fix: support export auth export with envvar for backwards compat

bharathkkb:
fixes https://github.com/google-github-actions/auth/issues/73

### DIFF
--- a/.github/workflows/setup-gcloud-it.yml
+++ b/.github/workflows/setup-gcloud-it.yml
@@ -269,17 +269,17 @@ jobs:
         - 'macos-latest'
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: 'actions/checkout@v2'
 
       - uses: 'actions/setup-node@v2'
         with:
           node-version: '12.x'
 
-      - id: build
-        name: Build dist
+      - id: 'build'
+        name: 'Build dist'
         run: 'npm ci && npm run build'
 
-      - uses: google-github-actions/auth@main
+      - uses: 'google-github-actions/auth@main'
         with:
           credentials_json: '${{ secrets.SETUP_GCLOUD_IT_KEY }}'
 
--- a/.github/workflows/setup-gcloud-it.yml
+++ b/.github/workflows/setup-gcloud-it.yml
@@ -277,13 +277,11 @@ jobs:
 
       - id: build
         name: Build dist
-        run: |-
-          npm ci
-          npm run build
+        run: 'npm ci && npm run build'
 
       - uses: google-github-actions/auth@main
         with:
-          credentials_json: ${{ secrets.SETUP_GCLOUD_IT_KEY }}
+          credentials_json: '${{ secrets.SETUP_GCLOUD_IT_KEY }}'
 
       - name: 'setup-gcloud'
         uses: './'
--- a/.github/workflows/setup-gcloud-it.yml
+++ b/.github/workflows/setup-gcloud-it.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
     - 'master'
-    - 'fix-export'
   pull_request:
     branches:
     - 'master'
@@ -259,7 +258,7 @@ jobs:
 
   credentials_json_exported:
     if: ${{ github.event_name == 'push' || github.repository == github.event.pull_request.head.repo.full_name && github.actor != 'dependabot[bot]' }}
-    name: 'auth action wtih key and setup-gcloud export'
+    name: 'auth action with json key and setup-gcloud export'
     runs-on: '${{ matrix.os }}'
     strategy:
       fail-fast: false
--- a/.github/workflows/setup-gcloud-it.yml
+++ b/.github/workflows/setup-gcloud-it.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
     - 'master'
+    - 'fix-export'
   pull_request:
     branches:
     - 'master'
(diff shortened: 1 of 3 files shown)


.github/workflows/setup-gcloud-it.yml
```diff
@@ -255,6 +255,46 @@ jobs:
...
+
+      - id: build
+        name: Build dist
+        run: |-
+          npm ci
+          npm run build
+
+      - uses: google-github-actions/auth@main
+        with:
+          credentials_json: ${{ secrets.SETUP_GCLOUD_IT_KEY }}
```
sethvargo, line 286: Nit quote

.github/workflows/setup-gcloud-it.yml
```diff
@@ -255,6 +255,46 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: 'actions/setup-node@v2'
+        with:
+          node-version: '12.x'
+
+      - id: build
+        name: Build dist
+        run: |-
```
sethvargo, line 280: Nit we can oneline this as `npm ci && npm run build`

src/setup-gcloud.ts
```diff
@@ -100,7 +101,26 @@ export async function run(): Promise<void> {
...
+      let serviceAccountKeyObj;
+      // if explicit SA passed, use that
+      if (serviceAccountKey) {
+        serviceAccountKeyObj = parseServiceAccountKey(serviceAccountKey);
+      } else if (process.env.GOOGLE_GHA_CREDS_PATH) {
+        const cPath = await fs.readFile(
+          process.env.GOOGLE_GHA_CREDS_PATH,
+          'utf8',
+        );
+        serviceAccountKeyObj = parseServiceAccountKey(cPath);
```
sethvargo, line 113: I don't think this will work, because the export from `auth` could be a WIF cred file, which isn't a "service account key"

sethvargo (changes requested):

kornbik (changes requested):
request a change

.github/workflows/setup-gcloud-it.yml
bharathkkb: done

.github/workflows/setup-gcloud-it.yml
bharathkkb: done

src/setup-gcloud.ts
bharathkkb: I refactored the logic a bit based on what we discussed yesterday and added some test cases for more scenarios. I also added a warning to help users migrating that export_creds is not necessary when using setup-gcloud in conjunction with auth.

src/setup-gcloud.ts
```diff
@@ -100,17 +100,33 @@ export async function run(): Promise<void> {
...
+        );
+        core.exportVariable('GCLOUD_PROJECT', serviceAccountKeyObj.project_id);
+      } else if (process.env.GOOGLE_GHA_CREDS_PATH) {
+        // No explicit SA input but process.env.GOOGLE_GHA_CREDS_PATH is set.
+        // process.env.GOOGLE_GHA_CREDS_PATH already contains credentials written to disk,
+        // so set credsPath to the existing cred filepath.
+        credsPath = process.env.GOOGLE_GHA_CREDS_PATH;
+        // User is likely using google-github-actions/auth for auth followed by setup-gcloud with export_default_credentials.
+        // This is unnecessary as auth already exports credentials.
+        core.warning(
```
sethvargo, line 118: `setup-gcloud` also exports `GOOGLE_GHA_CREDS_PATH` (maybe we should remove that?), so I think you'd hit this block with back-to-back `export_default_credentials` for two setup-gcloud uses.

We should add an integration test to make sure back-to-back setup-gcloud works and uses the correct auth:

```yaml
- uses: setup-gcloud
  with:
    credentials: cred1

- run: gcloud auth list | gregp 'cred1email'

- uses: setup-gcloud
  with:
    credentials: cred2

- run: gcloud auth list | gregp 'cred2email'
```

sethvargo: I _think_ it'll be fine because it'll fall into the first case of this conditional, but it gets weird if the user does:

```yaml
- uses: setup-gcloud
  with:
    credentials: cred1

- run: gcloud auth list | gregp 'cred1email'

# I think this will throw the warning
- uses: setup-gcloud
```

src/setup-gcloud.ts
```diff
@@ -100,17 +100,33 @@ export async function run(): Promise<void> {
...
-        JSON.stringify(serviceAccountKeyObj, null, 2), // Print to file as string w/ indents
-      );
+      // If explicit SA input, parse it, write to disk and set GCLOUD_PROJECT.
+      if (serviceAccountKey) {
+        const serviceAccountKeyObj = parseServiceAccountKey(serviceAccountKey);
+        await writeSecureFile(
+          credsPath,
+          JSON.stringify(serviceAccountKeyObj, null, 2), // Print to file as string w/ indents
+        );
+        core.exportVariable('GCLOUD_PROJECT', serviceAccountKeyObj.project_id);
```
sethvargo, line 110: I think we should only set the project ID here if the `project_id` input wasn't given above. If the user explicitly set a project_id, we should prefer that over the SA's project. Probably be good to compare the project IDs and throw a warning if they are different too

src/setup-gcloud.ts
```diff
@@ -100,17 +100,33 @@ export async function run(): Promise<void> {
...
+      } else if (process.env.GOOGLE_GHA_CREDS_PATH) {
```
sethvargo, line 111: Thinking out loud, why not just throw an exception if `export_default_credentials` is given without `service_account_key`?

src/setup-gcloud.ts
bharathkkb: Done and added a warning. In the earlier flow we were setting `GCLOUD_PROJECT` to `serviceAccountKeyObj.project_id` and then re exporting `GCLOUD_PROJECT` as `projectID` if set. Hopefully the newer flow is clearer.

src/setup-gcloud.ts
bharathkkb: Yeah the first case would be fine 

For
```yaml
- uses: setup-gcloud
  with:
    credentials: cred1
    
- uses: setup-gcloud
``` 
since user is not specifying explicit creds, it is no op.
Using something else other than `GOOGLE_GHA_CREDS_PATH` for cleanup in setup-gcloud is also feasible.

src/setup-gcloud.ts
bharathkkb: We could. My reason behind a warning rather than exception was to not break existing workflows if they had `export_default_credentials` when using with auth like https://github.com/google-github-actions/auth/issues/73

sethvargo: Makes sense. Can we update the warning text to say "this will be an error in a future release" and then when we cut the next major release we'll make it an error?

sethvargo (approved):

src/setup-gcloud.ts
bharathkkb: added

sethvargo (approved):

kaominghsu:
Hello,
I think the private_key of service_account passes through base64, resulting in insufficient length

sethvargo:
Hi @kaominghsu - if you're having an issue, please open a new GitHub issue and complete the issue template so we have the necessary information to debug.

PR #447, opened 2021-12-22
chore: update branch name for release workflow

bharathkkb:
We haven't switched from master to main in this repo yet and this is blocking build. I will create a follow up PR once I rename the branch.


sethvargo (approved):

sethvargo:
Yea, we should rename the branch. GitHub handles the redirects from master -> main, even for actions.

PR #469, opened 2022-01-05
Reword warning about HEAD

jsoref:
I received this warning because I was using this action.

I didn't use `@HEAD`, which would be pinning the action to `@HEAD`. It was *effectively* pinned to `HEAD` by using `@master`.

This change clarifies the language.

### DIFF
--- a/src/setup-gcloud.ts
+++ b/src/setup-gcloud.ts
@@ -36,7 +36,7 @@ const actionRef = process.env.GITHUB_ACTION_REF;
 export async function run(): Promise<void> {
   if (actionRef == 'main' || actionRef == 'master') {
     core.warning(
-      `google-github-actions/setup-gcloud is pinned at HEAD. We strongly ` +
+      `google-github-actions/setup-gcloud is effectively pinned to HEAD. We strongly ` +
         `advise against pinning to "@master" as it may be unstable. Please ` +
         `update your GitHub Action YAML from:\n\n` +
         `    uses: 'google-github-actions/setup-gcloud@master'\n\n` +


src/setup-gcloud.ts
```diff
@@ -36,7 +36,7 @@ const actionRef = process.env.GITHUB_ACTION_REF;
 export async function run(): Promise<void> {
   if (actionRef == 'main' || actionRef == 'master') {
     core.warning(
-      `google-github-actions/setup-gcloud is pinned at HEAD. We strongly ` +
+      `google-github-actions/setup-gcloud is effectively pinned to HEAD. We strongly ` +
```
sethvargo, line 39: We should probably just say "@master" instead here.

jsoref: You can't since line 37 also covers `main`.

This was the smallest change I could make that was definitely not wrong.

I'm not claiming that the line below that says `@master` is right...

sethvargo: Yea, but we know this particular action is pinned `@master` right now 😄 . We actually centralized this into a helper in https://github.com/google-github-actions/actions-utils/blob/f2e0cec9ce45d80c987943bdf3f5216734fa1f06/src/warnings.ts#L41, but there's some refactoring to do before we can bring that package into setup-gcloud right now.

If you wanted, you could grab the actual ref from `process.env.GITHUB_ACTION_REF`, which would theoretically handle either master or main.

jsoref: If you're ok with that change, I might be able to do it tonight.

sethvargo: SGMT!

jsoref:
Hmm, did I miss my chance to update? (Sorry, I have lots of things in the air)

sethvargo:
All good - we moved this into a different package to centralize the warning, so this became outdated

PR #522, opened 2022-03-01
chore: update dependabot settings

averikitsch:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,10 +7,14 @@ version: 2
 updates:
   - package-ecosystem: "npm"
     directory: "/"
-    schedule:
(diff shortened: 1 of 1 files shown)


sethvargo (approved):

craigdbarber:
Could be that some of the Windows VMs are now configured using Python3.9 by default? If so, this might be a more widespread issue.

bharathkkb:
Newer versions of gcloud do support 3.9, I think we were just using a really old version in the versioned tests which was causing this https://github.com/google-github-actions/setup-gcloud/pull/524/commits/ca069becf3150aff4d888f9dd6a29bae59c99ab4

sethvargo (approved):

PR #537, opened 2022-03-20
feat: throw a very specific error when pinned to master

sethvargo:
This can be overridden by setting an environment variable.

### DIFF
--- a/src/setup-gcloud.ts
+++ b/src/setup-gcloud.ts
@@ -40,7 +40,44 @@ export const GCLOUD_METRICS_LABEL = 'github-actions-setup-gcloud';
 export async function run(): Promise<void> {
   // Warn if pinned to HEAD
   if (isPinnedToHead()) {
-    core.warning(pinnedToHeadWarning('v0'));
+    // TODO(sethvargo): switch back to warning level after branch is renamed
+    // from "master" -> "main".
+    core.error(pinnedToHeadWarning('v0'));
+
+    // TODO(sethvargo): remove after branch is renamed from "master" -> "main".
+    const envvar =
+      'SETUP_GCLOUD_I_UNDERSTAND_USING_MASTER_WILL_BREAK_MY_WORKFLOW_ON_2022_04_05';
+    const issueURL =
+      'https://github.com/google-github-actions/setup-gcloud/issues/539';
+    const isMaster = process.env.GITHUB_ACTION_REF === 'master';
+    const isOurs =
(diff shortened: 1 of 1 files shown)


src/setup-gcloud.ts
```diff
@@ -40,7 +40,36 @@ export const GCLOUD_METRICS_LABEL = 'github-actions-setup-gcloud';
...
-    core.warning(pinnedToHeadWarning('v0'));
+    // TODO(sethvargo): switch back to warning level after branch is renamed
+    // from "master" -> "main".
+    core.error(pinnedToHeadWarning('v0'));
+
+    // TODO(sethvargo): remove after branch is renamed from "master" -> "main".
+    const envvar =
+      'SETUP_GCLOUD_I_UNDERSTAND_USING_MASTER_WILL_BREAK_MY_WORKFLOW_ON_APRIL_5_2022';
+    if (process.env.GITHUB_ACTION_REF === 'master' && !process.env[envvar]) {
+      throw new Error(
```
sethvargo, line 51: It's worth noting that this will break all of _our_ workflows too (especially ones in other repositories). We intentionally float our integration tests to HEAD so we catch breaking changes early.

bharathkkb: Could we check for GITHUB_REPOSITORY_OWNER to prevent this?

src/setup-gcloud.ts
```diff
@@ -40,7 +40,36 @@ export const GCLOUD_METRICS_LABEL = 'github-actions-setup-gcloud';
...
+          `    uses: 'google-github-actions/setup-gcloud@master'\n` +
+          `\n` +
+          `to:\n` +
+          `\n` +
+          `    uses: 'google-github-actions/setup-gcloud@v0'\n` +
+          `\n` +
+          `While not recommended, you can still pin to the "master" branch ` +
+          `by setting the environment variable "${envvar}=1". However, on ` +
+          `2022-04-05 when the branch is renamed, all your workflows will ` +
+          `begin failing with an obtuse and difficult to diagnose error ` +
```
bharathkkb, line 69: Lets also open a pinned issue before rename with the error before we rename so users know what to do.

bharathkkb (approved):

src/setup-gcloud.ts
sethvargo: https://github.com/google-github-actions/setup-gcloud/issues/539

PR #541, opened 2022-03-21
chore: setFailed instead of throwing

sethvargo:


src/setup-gcloud.ts
```diff
@@ -77,6 +75,7 @@ export async function run(): Promise<void> {
           `\n` +
           `For more information, please see ${issueURL}.`,
```
bharathkkb, line 76: Issue URL click through was also broken due to the period. https://github.com/bharathkkb/sample-app/runs/5635742655?check_suite_focus=true#step:3:31

sethvargo: Fixed.

bharathkkb (approved):

PR #582, opened 2022-11-08
fix: cleanup integration tests

verbanicm:

### DIFF
--- a/.github/workflows/setup-gcloud-it.yml
+++ b/.github/workflows/setup-gcloud-it.yml
@@ -17,7 +17,6 @@ jobs:
   versioned:
     if: ${{ github.event_name == 'push' || github.repository == github.event.pull_request.head.repo.full_name && github.actor != 'dependabot[bot]' }}
     name: setup-gcloud versioned
-    runs-on: ${{ matrix.operating-system }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/setup-gcloud-it.yml
```diff
@@ -17,7 +17,7 @@ jobs:
   versioned:
     if: ${{ github.event_name == 'push' || github.repository == github.event.pull_request.head.repo.full_name && github.actor != 'dependabot[bot]' }}
     name: setup-gcloud versioned
-    runs-on: ${{ matrix.operating-system }}
+    runs-on: ${{ matrix.os }}
```
sethvargo, line 20: Can you move this below the matrix everywhere?

.github/workflows/setup-gcloud-it.yml
```diff
@@ -396,8 +394,6 @@ jobs:
         uses: ./
         with:
           version: 'latest'
-          service_account_email: ${{ secrets.SETUP_GCLOUD_IT_EMAIL }}
```
sethvargo, line 399: Why remove that?

verbanicm: it wasnt being used in the code anywhere

.github/workflows/setup-gcloud-it.yml
verbanicm: updated

sethvargo (approved):

PR #584, opened 2022-11-09
fix: suggest v1 instead of v0

KengoTODA:
Thanks for releasing v1, it encourages me to keep using fresh releases!

Here I'll suggest a minor fix, which helps users to use the latest release instead of v0.


verbanicm (approved):

verbanicm:
Thanks!!

PR #586, opened 2022-11-10
Suggestion to fix missing GCloud SDK authentication

KengoTODA:
Hello 👋 

I'm trying to bump up `google-github-actions/setup-gcloud` to v1, and found that there is one deleted feature that was not deprecated in the v0 README.

## Expected Behavior (v0)

v0 authenticates GCloud SDK if `GOOGLE_GHA_CREDS_PATH` env var is set. `google-github-actions/auth` sets this env var (I've checked both v1 and v0), so we use these actions like below:

```yaml
    steps:
    - id: 'auth'
      uses: 'google-github-actions/auth@v1'
      with:
        workload_identity_provider: 'projects/123456789/locations/global/workloadIdentityPools/my-pool/providers/my-provider'
        service_account: 'my-service-account@my-project.iam.gserviceaccount.com'

    - name: 'Set up Cloud SDK' # ⚠ does not work with v1
      uses: 'google-github-actions/setup-gcloud@v0'

    - run: 'gsutil rsync ...'
```

## Actual Behavior (v1.0.0)

[According to the changeset of v1.0.0](https://github.com/google-github-actions/setup-gcloud/compare/v0...v1.0.0#diff-7887ee7d4ed66a4199b461b9af71c4ef755792fa707724144fdf30a03c36070bL85-L89), v1.0.0 release removed this automatic authentication feature.

I haven't confirmed yet, but we probably need a snippet like below to run authentication manually:

```yaml
   - run: gcloud --quiet auth activate-service-account "my-service-account@my-project.iam.gserviceaccount.com" --key-file "$GOOGLE_GHA_CREDS_PATH"
```

## Discussion I want to hold here

I hope we can bring this automatic authentication feature back in v1 release. Could you consider it when you have time?

Or if it is necessary to remove this code, I hope we users have enough document in README.md or somewhere, to find how we need to authenticate GCloud SDK.

Thanks for reading my PR! 🙌


KengoTODA:
Thanks for quick reaction! 😀

PR #624, opened 2023-03-24
Switch to pull non-secret values from env

sethvargo:


verbanicm (approved):

rtrzebinski:
isn't README.md outdated now still mentioning 'secrets'?

sethvargo:
You can use secrets or variables. We have examples using secrets in the readme because it's possible for project-ids to contain PII or secret information.

rtrzebinski:
I understand, but docs should math the example, otherwise it won't work.

rtrzebinski:
I mean README.md here in the repo.

sethvargo:
I still don't understand what you're talking about. This PR switches to use non-secrets for data that we consider to be non-secret. For example, we do not consider Google Cloud Project IDs to be a secret in this project because it doesn't give away any information. Since Project IDs are named by end-users, they _could_ contain PII and therefore might be stored as secrets or variables. The resolution of secrets and variables is handled by GitHub, not us; it gets interpolated into the YAML.

The README references `secrets.GCP_CREDENTIALS`, which _is_ a secret and remains a secret.

Again, I'm not really sure I understand what you're saying.

rtrzebinski:
In the following PR:

https://github.com/google-github-actions/setup-gcloud/pull/624/files

In file:

example-workflows/gke/.github/workflows/gke.yml

The change:

```
-  PROJECT_ID: ${{ secrets.GKE_PROJECT }}
+ PROJECT_ID: ${{ vars.GKE_PROJECT }}
```

Is not reflected in file:

https://github.com/google-github-actions/setup-gcloud/blob/main/example-workflows/gke/README.md

Which is still saying:

```
Add the following secrets to your repository's secrets:

GKE_PROJECT: Google Cloud project ID
```

While it is not a secret anymore, but variable.

Issue #594, opened 2022-11-19
Authenticaton warning when using application default credentials


jorob16:
### TL;DR

I'm running a self-hosted GitHub Actions runner on a GKE cluster. Permissions are handled through application default credentials (Workload Identity) and my workflow works correctly (no authorization errors). Unfortunately `setup-gcloud` still spits out a warning: `**Warning**: No authentication found for gcloud, authenticate with `google-github-actions/auth`.`

![afbeelding](https://user-images.githubusercontent.com/91264722/202873095-d1124aee-0308-4153-85b6-c91b466c182a.png)
![afbeelding](https://user-images.githubusercontent.com/91264722/202873357-1111b074-ada1-441c-a07d-1c6997b8c2b2.png)

### Expected behavior

I did not expect a warning, considering that the workflow works correctly and interacts with multiple GCP services for which authorization is definitely required.

### Observed behavior

An authentication warning is displayed in the workflow logs and as an annotation under the job overview.

### Action YAML

```yaml
name: CICD Bumpversion master
on: 
  push:
    branches:
      - master
    paths:
      - cicd-bumpversion/**
      - .github/workflows/cicd-bumpversion-master.yaml
jobs:
  build-image:
    name: Build and push image
    runs-on: [self-hosted, **REDACTED**]
    steps:
      - name: Set up Cloud SDK
        uses: google-github-actions/setup-gcloud@v1
      - name: Checkout
        uses: actions/checkout@v3
      - name: Build Docker image
        run: docker build -t **REDACTED**/cicd-bumpversion:latest .
        working-directory: cicd-bumpversion/
      - name: Push image to Container Registry
        run: |
          gcloud auth configure-docker
          docker push **REDACTED**/cicd-bumpversion:latest
```

sethvargo:
Hi @jorob16 

This is correct and this is just a warning. We emit a warning when there are no credentials set by the `auth` action. This is just a warning and can be safely ignored for your use case.

kkroening:
> this is just a warning

Here's what "just a warning" looks like for anyone hoping to use `setup-gcloud` in CI - e.g. to `install-components: cloud-datastore-emulator` for testing:

<img src="https://user-images.githubusercontent.com/2035953/207767108-fee0f6c4-7cfc-4fc4-9db2-fca66c19310c.png" alt="Warnings" width="65%"/>

[...]

PR #633, opened 2023-05-04
Check if authenticated before trying to auth

isarkis:
Addresses #594 by not throwing 'No authentication' warning when using application default credentials.


sethvargo:
This won't work. There are multiple use cases for running `auth` and `setup-gcloud` multiple times in the same workflow. Nothing is "thrown" because this is not an error, it's a warning.

PR #696, opened 2024-10-11
feat: Added optional inputs 

mohsin:
Added the optional inputs: workload_identity_provider, service_account, credentials_json

Without these declared in action.yml, Github runners will keep throwing "Unexpected input(s)" warnings whenever the action is run.

Screenshot 2024-10-12 at 02 03 47
Screenshot 2024-10-12 at 02 03 32

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -57,6 +57,28 @@ inputs:
       at other inputs and environment variables.
     required: false
 
+  workload_identity_provider:
+    description: |-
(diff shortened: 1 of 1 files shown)


sethvargo:
That's because those are not valid inputs to setup-gcloud. In the examples, those inputs are in `auth`.

mohsin:
Oh, okay

PR #736, opened 2026-06-24
feat: add version_file input to read the gcloud version from a file

somaz94:
Adds an optional `version_file` input that resolves the Cloud SDK version from a file, mirroring the `*-version-file` pattern in `actions/setup-go`, `actions/setup-node`, and `actions/setup-python`.

Two formats are supported:
- a plain version file containing a single version string or constraint (e.g. `416.0.0` or `>= 416.0.0`)
- an asdf/mise `.tool-versions` file containing a `gcloud <version>` line

Precedence: an explicitly set `version` (anything other than the default `latest`) wins and `version_file` is ignored with a warning. The version-resolution logic is unchanged once a concrete version is determined, so the existing install/cache path is reused as-is.

Validation (run locally):
- `npm run lint` (eslint) passes
- `npm test` passes, 22/22 (added 3 `run()` cases for `.tool-versions` / plain file / `version` precedence, plus 5 `parseVersionFile` unit cases)
- `npm run build` (ncc) rebuilt `dist/index.js`
- `npm run docs` regenerated the README inputs section


somaz94:
Friendly ping — this has been green and unreviewed for about a month.

One clarification in case the bot comment above caused confusion: the CLA **is** signed — the `cla/google` check passes on the current head. The 2026-06-24 "you will need to sign" comment is stale.

The change adds a `version_file` input so the gcloud version can be read from a file instead of being pinned inline. Happy to adjust if the input name or lookup order should differ.

somaz94:
Closing this — it has been open since June with no review, and I would rather not leave a stale PR sitting in the queue. Happy to reopen if the version_file input is something you would like.

